PR #9, opened 2017-09-03
Fix make clean and build instructions

sadikovi:
This PR adds build instructions for the project. I have spent a considerable amount of time trying to build Thrift and etc., so I hope this will be useful for other people.

Also `make clean` is updated to remove generated files from `src/parquet_thrift`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,14 @@ An [Apache Parquet](https://parquet.apache.org/) implementation in Rust (work in
 - Rust nightly
 - Thrift 0.11+ or master (with Rust support)
 
-To install Rust nightly build, just run (assuming you have Rust installed already):
+To install Rust nightly build, run this (assuming you have Rust installed already):
 ```shell
 rustup update nightly
 rustup default nightly
 ```
 
 Project requires Thrift with Rust support, which is in 0.11+/master. If release is not available on
-official website, follow instructions:
+[official website](https://thrift.apache.org), build from source:
 ```shell
 git clone --depth=1 https://github.com/apache/thrift
 # go to thrift project directory and install it
@@ -31,21 +31,23 @@ PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig:"${PKG_CONFIG_PATH}"
 export LD_LIBRARY_PATH CPATH PKG_CONFIG_PATH
 
 ./bootstrap.sh
-# when configuring you might want to keep set of languages to a minimum, e.g. C++/Java/Rust
-./configure
+# when configuring you might want to keep set of languages to a minimum, e.g. C++/Rust
+# and disable the rest with '--with-LANG=no', see './configure --help'
+./configure --enable-tests=no --enable-tutorial=no
 
 make install
 ```
 
 On OS X you might need to pre-install following packages (if not installed already):
 ```shell
-brew install openssl libtool automake pkg-config
+brew install boost libevent openssl libtool automake pkg-config
 ```
 
 ### Build and test
 To build project run:
 ```shell
 cd parquet-rs
+
 make
 
 # clean generated files
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ An [Apache Parquet](https://parquet.apache.org/) implementation in Rust (work in
 ## Development
 ### Requirements
 - Rust nightly
-- Thrift 0.11+ or master (with Rust support)
+- Thrift (with Rust support)
 
 To install Rust nightly build, run this (assuming you have Rust installed already):
 ```shell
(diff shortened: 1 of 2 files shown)


sadikovi:
@sunchao Could you review this PR? I would like to help with this project, if possible (though I am new to Rust). Thanks.

README.md
```diff
@@ -5,3 +5,55 @@
...
+
+# export env vars for openssl
+LD_LIBRARY_PATH=/usr/local/opt/openssl/lib:"${LD_LIBRARY_PATH}"
+CPATH=/usr/local/opt/openssl/include:"${CPATH}"
+PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig:"${PKG_CONFIG_PATH}"
+export LD_LIBRARY_PATH CPATH PKG_CONFIG_PATH
+
+./bootstrap.sh
+# when configuring you might want to keep set of languages to a minimum, e.g. C++/Java/Rust
+./configure
```
sunchao: Perhaps we can add `enable-libs=no` here to avoid building all the language libs that we don't need?

sunchao:
Thanks @sadikovi for your interest in this project. PRs are definitely welcome! I left a minor comment but overall the code looks good.

README.md
sadikovi: I just tried doing it, and it disables all libraries, e.g. `./configure --enable-libs=no --with-rust=yes`, all language options are no-op. I guees it is because of this https://github.com/apache/thrift/blob/master/configure.ac#L116. It looks like one would have to disable them manually. 

I updated README to include flags for no tests and tutorials. Also found that Python and PHP Thrift installations do not work (require additional privileges), but I did not mention it in README.

I can be wrong here, would appreciate if you could help with `enable-libs`.

sadikovi:
@sunchao Thanks for reviewing PR. I answered your question regarding `enable-libs` - honestly, do not know how to disable them all and enable certain ones - would appreciate your help.

Also do you think I should change titles in README, e.g. `Development` and etc.? Thanks.

README.md
sunchao: Hmm.. did you get any error with `./configure --enable-libs=no --with-rust=yes`?: I think it should be `with-rs=yes`. Also, I don't think we need to have `with-rs=yes` since `cargo build` will build this dependency (see [here](https://github.com/sunchao/parquet-rs/blob/master/Cargo.toml#L21)). This is also [how the CI does today](https://github.com/sunchao/parquet-rs/blob/master/.travis.yml#L18).

README.md
```diff
@@ -11,14 +11,14 @@ An [Apache Parquet](https://parquet.apache.org/) implementation in Rust (work in
...
 - Thrift 0.11+ or master (with Rust support)
 
-To install Rust nightly build, just run (assuming you have Rust installed already):
+To install Rust nightly build, run this (assuming you have Rust installed already):
 ```shell
 rustup update nightly
 rustup default nightly
 ```
 
 Project requires Thrift with Rust support, which is in 0.11+/master. If release is not available on
```
sunchao: I think we can change "0.11+/master" to "master": I'm not sure if a 0.11 release will ever happen. And we can change this after the next version is released.

sadikovi: Okay, will change.

sunchao:
> Also do you think I should change titles in README, e.g. Development and etc.? Thanks.

Yes a `Development` section is fine to me.

README.md
sadikovi: I ran command `./configure --enable-libs=no --with-rs` and it listed libraries for each language - they all had "no" including Rust. Is this okay?

sadikovi: I get something like this:
```
./configure --enable-libs=no --with-rs
...

thrift 1.0.0-dev

Building Plugin Support ...... : no
Building C++ Library ......... : no
Building C (GLib) Library .... : no
Building Java Library ........ : no
Building C# Library .......... : no
Building .NET Core Library ... : no
Building Python Library ...... : no
Building Ruby Library ........ : no
Building Haxe Library ........ : no
Building Haskell Library ..... : no
Building Perl Library ........ : no
Building PHP Library ......... : no
Building Dart Library ........ : no
Building Erlang Library ...... : no
Building Go Library .......... : no
Building D Library ........... : no
Building NodeJS Library ...... : no
Building Lua Library ......... : no
Building Rust Library ........ : no
```

sadikovi: I am running `make install` to see what happens.

sadikovi: Yes, you are right, just running `./configure --enable-libs=no` works! Thank you! I will update the doc.

sadikovi:
@sunchao I updated README file. Could you have a look again? Thanks!

sunchao:
Merged. Thanks!

sadikovi:
Thanks for merging pull request!

PR #12, opened 2017-09-05
Fix reading required non-repeated column

### LINKED ISSUE #11: Issue when reading required INT32 column with PLAIN encoding
For required fields with PLAIN encoding, current code `read_batch` in `ColumnReaderImpl` goes into infinite loop. It happens when field is required, it does not have repetition/definition levels, therefore `values_to_read` keeps being set to 0, which results in buffer being `&[values_read..values_read + values_to_read]` -> empty. This goes into infinite loop because decoding for empty buffer returns 0, and loop never progresses.

Fix for the issue is setting `values_to_read` explicitly, when field is required (assertion is optional):
```diff
diff --git a/src/column/reader.rs b/src/column/reader.rs
index 9d36918..2a230e8 100644
--- a/src/column/reader.rs
+++ b/src/column/reader.rs
@@ -154,6 +154,8 @@ impl<'a, T: DataType> ColumnReaderImpl<'a, T> where T: 'static {
             }
           }
         }
+      } else {
+        values_to_read = batch_size;
       }
 
       if self.descr.max_rep_level() > 0 && rep_levels.is_some() {
diff --git a/src/encodings/decoding.rs b/src/encodings/decoding.rs
index cf98276..a76ace8 100644
--- a/src/encodings/decoding.rs
+++ b/src/encodings/decoding.rs
@@ -127,6 +127,7 @@ default impl<T: DataType> Decoder<T> for PlainDecoder<T> {
   #[inline]
   fn get(&mut self, buffer: &mut [T::T]) -> Result<usize> {
     assert!(self.data.is_some());
+    assert!(buffer.len() != 0);
 
     let data = self.data.as_mut().unwrap();
     let num_values = cmp::min(buffer.len(), self.num_values);
```

Here is schema information for which it fails.
```
version: 1
num of rows: 4
created by: parquet-mr version 1.8.1 (build 4aba4dae7bb0d4edbcf7923ae1339f28fd3f7fcf)
message spark_schema {
  REQUIRED INT32 a;
  REQUIRED INT64 b;
  OPTIONAL BYTE_ARRAY c;
  REQUIRED group d {
    REQUIRED INT32 a;
    REQUIRED INT64 b;
    OPTIONAL BYTE_ARRAY c;
  }
  REQUIRED group e (LIST) {
    REPEATED group list {
      REQUIRED INT32 element;
    }
  }
}
```

sadikovi:
This PR fixes issue when column is required and non-repeated, for example, INT32 column `a` in schema below:
```
message spark_schema {
  REQUIRED INT32 a;
  REQUIRED INT64 b;
  OPTIONAL BYTE_ARRAY c;
  REQUIRED group d {
    REQUIRED INT32 a;
    REQUIRED INT64 b;
    OPTIONAL BYTE_ARRAY c;
  }
  REQUIRED group e (LIST) {
    REPEATED group list {
      REQUIRED INT32 element;
    }
  }
}
```

Before patch, when reading such column code resulted in passing empty buffer into decoder, which would return 0 values read, which would cause loop never exit.

After this patch, we assign full batch size as approximate values to read, when non-repeated column. This value is still adjusted later based on actual values in page.

Added couple of unit tests similar to existing ones, also tested manually on a Parquet file.

Fixes #11.


src/column/reader.rs
```diff
@@ -671,8 +701,11 @@ mod tests {
             num_values_cur_page += 1;
           }
         }
-      } else {
+      } else if levels_per_page > 0 {
         num_values_cur_page = levels_per_page;
+      } else {
+        // for testing required non-repeated columns (levels_per_page == 0)
+        num_values_cur_page = batch_size;
```
sadikovi: It looks like `num_values_cur_page` was only set when levels per page are defined. In my case I set it explicitly to 0, therefore I have to update number of values in page to propagated `batch_size`. Not sure, if it is a good way of doing it.

sadikovi:
@sunchao please review, tests reproduce problem when fix is commented out, and pass otherwise!

src/column/reader.rs
sunchao: It seems your tests still use `NUM_LEVELS` instead of 0? Also `batch_size` is not added to the `make_pages` function so maybe the build won't pass.

Also I think there's another stupid bug of mine in the following that may be related:
```rust
let mut page_builder = DataPageBuilderImpl::new(desc.clone());
if max_def_level > 0 {
  page_builder.add_rep_levels(max_rep_level, &rep_levels[level_range.clone()]);
}
if max_rep_level > 0 {
  page_builder.add_def_levels(max_def_level, &def_levels[level_range]);
}
```
It should process def levels in the first if and rep levels in the second if.

sadikovi: Yes, you are right. I think it is still uses NUM_LEVELS to get values per page, not batch size. I will have a closer look at `make_pages` function and will also update code with your changes, and will try to come up with unittest for it. 

I am super confused with definition and repetition levels at this point:) Need to spend some time figuring out which is which.

sadikovi:
@sunchao I updated code with your suggestions, also fixed code in `make_pages` function that you pointed out. Could you review it again? Thanks.

src/column/reader.rs
```diff
@@ -615,8 +642,10 @@ mod tests {
     }
 
     fn add_values<T: DataType>(&mut self, encoding: Encoding, values: &[T::T]) where T: 'static {
-      assert!(self.num_values >= values.len() as u32, "num_values: {}, values.len(): {}",
-              self.num_values, values.len());
+      // when column is non-repeated, we should set all values in slice
+      if self.num_values < values.len() as u32 {
```
sunchao: Seems the `num_values` is set in `add_rep_levels`, which is problematic when `rep_level` is 0. Maybe we should pass the value for this from outside, perhaps using `levels_per_page`. What do you think?

sunchao:
@sadikovi Thanks this looks much better! Just have one more comment :)

src/column/reader.rs
sadikovi: My reasoning was that we should not set num values in add_rep_levels function, but I also feel bad about removing the assertion in add_values method:). 

We should definitely keep assert. Could you elaborate a bit on your suggestion?

sunchao: `levels_per_page` ([see here](https://github.com/sadikovi/parquet-rs/blob/5e3f2f54beb6093d8fd5ce67ac63c6a2ac7fdecf/src/column/reader.rs#L677)) is the same as the number of values in the page (including nulls) I think. So we can pass that to the [ctor](https://github.com/sadikovi/parquet-rs/blob/5e3f2f54beb6093d8fd5ce67ac63c6a2ac7fdecf/src/column/reader.rs#L614) of `DatePageBuilderImpl` and assign that to `num_values`. Let me know if this makes sense. :)

sadikovi: I think I understand. Will patch it, see what happens:)

sadikovi:
@sunchao I made code changes based on your suggestion. Added back assertion on `num_values`.

sunchao:
@sadikovi Thanks. LGTM.

sadikovi:
@sunchao thanks for review!

PR #14, opened 2017-09-08
Add usage examples

sadikovi:
This PR adds examples for extracting file metadata and reading data.

I added Parquet file that contains example data, such as primitive columns, struct fields and list fields to show different usage. It does not contain null values. Below the actual data:
```
message spark_schema {
  required int32 a;
  required int64 b;
  optional binary c (UTF8);
  required group d {
    required int32 a;
    required int64 b;
    optional binary c (UTF8);
  }
  required group e (LIST) {
    repeated group list {
      required int32 element;
    }
  }
}

+----+---+---+------------+------------+
|   a|  b|  c|           d|           e|
+----+---+---+------------+------------+
|   1|  2|abc|   [1,2,abc]|      [1, 1]|
|  10|  3|def|  [10,3,def]|    [10, 10]|
| 100|  4|ghi| [100,4,ghi]|  [100, 100]|
|1000|  4|jkl|[1000,4,jkl]|[1000, 1000]|
+----+---+---+------------+------------+
```

Also updated README with links to source files and commands to run.


sadikovi:
@sunchao could you have a look at this PR? 

This is a very first pass on usage examples - my attempts to read file with the library. Could you let me know if code is okay and/or I am doing something wrong reading a file?

Would like to know your opinion on adding actual code vs adding it into README directly and binary example Parquet file. 

Thanks!

src/bin/example-metadata.rs
```diff
@@ -0,0 +1,63 @@
...
+// Column INT32: "a"
+// Column INT64: "b"
+// Column BYTE_ARRAY: "c"
+// Column INT32: "d.a"
+// Column INT64: "d.b"
+// Column BYTE_ARRAY: "d.c"
+// Column INT32: "e.list.element"
+//
+fn main() {
+  let path = Path::new("/Users/sadikovi/developer/parquet-rs/data/example.snappy.parquet");
```
sunchao: This is not gonna work. Maybe change to a relative path?

src/bin/example-metadata.rs
```diff
@@ -0,0 +1,63 @@
...
+// Number of rows: 4
+// Column INT32: "a"
+// Column INT64: "b"
+// Column BYTE_ARRAY: "c"
+// Column INT32: "d.a"
+// Column INT64: "d.b"
+// Column BYTE_ARRAY: "d.c"
+// Column INT32: "e.list.element"
+//
+fn main() {
```
sunchao: I got:

```
➜  parquet-rs (example) ✔ cargo run --bin example-read
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/example-read`
Reading batch of records: 4
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NYI("Unsupported encoding PLAIN_DICTIONARY")', src/libcore/result.rs:860:4
```
while running this. Anything I missed?

src/bin/example-read.rs
```diff
@@ -0,0 +1,151 @@
...
+fn read_values<T: DataType>(
+    reader: &mut ColumnReaderImpl<T>,
+    batch_size: usize) -> (usize, Vec<T::T>, Vec<i16>, Vec<i16>) where T: 'static {
+
+  let mut values = vec![T::T::default(); batch_size];
+  let mut def_levels = vec![i16::default(); batch_size];
+  let mut rep_levels = vec![i16::default(); batch_size];
+  let mut curr_values_read = 0;
+  let mut curr_levels_read = 0;
+  loop {
```
sunchao: This loop doesn't seem to work if number of values is greater than `batch_size`.

src/bin/example-read.rs
```diff
@@ -0,0 +1,151 @@
...
+    curr_levels_read += levels_read;
+
+    if values_read == 0 {
+      break;
+    }
+  }
+  (curr_values_read, values, def_levels, rep_levels)
+}
+
+fn main() {
```
sunchao: Similar to `read-metadata`, I got this error:
```
➜  parquet-rs (example) ✗ cargo run --bin example-read
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/example-read`
Reading batch of records: 4
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NYI("Unsupported encoding PLAIN_DICTIONARY")', src/libcore/result.rs:860:4
```

README.md
```diff
@@ -6,6 +6,31 @@
 
 An [Apache Parquet](https://parquet.apache.org/) implementation in Rust (work in progress)
 
+## Example
+Library usage examples can be found in [src/bin](./src/bin) folder.
+
+### Reading file metadata
```
sunchao: Is it possible we can combine this with `dump-schema`?

src/bin/example-metadata.rs
sadikovi: I fixed it, thanks!

src/bin/example-read.rs
sadikovi: Fixed, thanks!
Now I pass total values to read. Also when changing batch_size in `main()`, this will read and parse data accordingly.

src/bin/example-metadata.rs
sadikovi: I am not sure. Did you check out my branch? It looks like the problem that I patched the other day. Could you try to fetch latest commits from master?

README.md
sadikovi: Yes, we definitely can. I will update README.

src/bin/example-read.rs
sadikovi: Could you try fetching latest commits from master? I think this is one of the problems that we solved recently. It works from my branch that contains those fixes.

sadikovi:
@sunchao I updated PR, could you review again?

sadikovi:
I think we should invest time to build high level read API, rather than merging this PR. Once that is done we could add small readme and publish API docs (scheme building and reading). Let me know what you think.

src/bin/example-read.rs
```diff
@@ -0,0 +1,153 @@
...
+            records[index].d_c = String::from_utf8(values[index].as_bytes().to_vec()).unwrap();
+          }
+        },
+        // repeated list field "e"
+        (Type::INT32, "e.list.element") => {
+          // we need to know max repetition level to reconstruct list, because values
+          // are stored back to back.
+          let max_rep_level = column.column_descr().max_rep_level();
+          let num_levels = max_rep_level as usize + 1;
+          // ... let max_def_level = column.column_descr().max_def_level();
```
sunchao: Remove?

src/bin/example-read.rs
```diff
@@ -0,0 +1,153 @@
...
+          // in practice this should follow Parquet repeated field assembly that can be found here:
+          // https://blog.twitter.com/engineering/en_us/a/2013/dremel-made-simple-with-parquet.html
+          // the basic idea is resolving repetition level 0, 1, 2, ...; 0 level means beginning of
+          // the record and all levels must be created up to max repetition level. Each number that
+          // is different from 0 indicates new value on that repetition level.
+          let mut assembly = Vec::new();
+          for k in 0..values_read {
+            if rep_levels[k] == 0 {
+              assembly.push(Vec::new());
+              assembly.last_mut().map(|v| v.push(values[k]));
```
sunchao: nit: can move this out of the if expr.

sadikovi:
@sunchao thanks for the review! I will address your comments and fix this in couple of days - need to learn more about parsing in parquet-cpp and parquet-mr.

sadikovi:
I am going to close this for now, will revisit later once I have patched several things.

PR #16, opened 2017-09-18
Support for DataPage V2

sadikovi:
This PR adds support for DataPage V2. 

I tested manually on a table that was written with version 2 writer. Refactored DataPage code into `set_current_page_encoding`, since it is the same for both V1 and V2. Also added method to set data range directly in RLE decoder, because DataPage V2 has repetition and definition levels separate from actual page data, which can be compressed.

It looks like DataPage V2 encodes int32 values with Delta Bit Packing. Currently decoder supports only int64, but I patched it manually - it the same for int32. Will create separate pull request for that.


sadikovi:
@sunchao I did not add unittests for this change in respect to reading DataPage V2. Can you advice on how I can write unittests to tests reading DataPage V2?

src/file/reader.rs
```diff
@@ -285,21 +285,45 @@ impl PageReader for SerializedPageReader {
...
+          let decompressed_size = decompressor.decompress(&buffer[offset..], &mut decompressed_buffer)?;
+          if decompressed_size != uncompressed_len {
+            return Err(general_err!("Actual decompressed size doesn't \
+              match the expected one ({} vs {})", decompressed_size, uncompressed_len));
+          }
+          if offset == 0 {
+            buffer = decompressed_buffer;
+          } else {
+            // prepend saved offsets to the buffer
+            buffer = buffer[..offset].to_vec();
```
sunchao: nit: replace by `truncate()`?

src/file/reader.rs
```diff
@@ -285,21 +285,45 @@ impl PageReader for SerializedPageReader {
   fn get_next_page(&mut self) -> Result<Option<Page>> {
     while self.seen_num_values < self.total_num_values {
       let page_header = self.read_page_header()?;
-      let compressed_len = page_header.compressed_page_size as usize;
-      let uncompressed_len = page_header.uncompressed_page_size as usize;
-      let mut buffer = vec![0; compressed_len];
+
+      // when processing data page v2, depending on enabled compression for the page, we should
+      // account for uncompressed data ('offset') of repetition and definition levels.
+      let (offset, can_apply_compression) = if page_header.data_page_header_v2.is_some() {
```
sunchao: nit: this can be replaced by `if let ...`. Also, maybe change `can_apply_compression` to something like `is_compressed` or `can_decompress`?

src/encodings/levels.rs
```diff
@@ -107,6 +107,14 @@ impl LevelDecoder {
   }
 
   #[inline]
+  // Set byte array explicitly when start position ('start') and length ('len') are known in advance
+  // Returns number of total bytes set for this decoder (len)
+  pub fn set_data_range(&mut self, data: &ByteBufferPtr, start: usize, len: usize) -> usize {
```
sunchao: nit: instead of having a separate function, you can also add a flag to `set_data` to indicate whether the data size should be read from the first i32 of the input buffer. The current solution is fine though so feel free to pick either one. :)

sunchao:
@sadikovi Thanks for the PR (as always)! 

> @sunchao I did not add unittests for this change in respect to reading DataPage V2. Can you advice on how I can write unittests to tests reading DataPage V2?

I would suggest to add two unit tests: one to test the `read_new_page` added to the column reader, and another for `get_next_page` in `file/reader`. 

For the former you may need to modify the existing `DataPageBuilder` and `LevelEncoder` (right now it always emit a 4-byte i32 at the beginning of buffer); for the latter, you can add a new Parquet test file under `/data` and then just do some simple sanity test like the `test_file_reader` in the file.

src/encodings/levels.rs
```diff
@@ -107,6 +107,14 @@ impl LevelDecoder {
   }
 
   #[inline]
+  // Set byte array explicitly when start position ('start') and length ('len') are known in advance
```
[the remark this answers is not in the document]
sadikovi: Yes, will do - this coded that way, because of me not understanding those attributes. I will update all similar occurrences.

Could you explain, what `#[inline]` does? Thanks!

src/encodings/levels.rs
sadikovi: Thanks for suggestion! Would it mean that I had to encode both start position and length? I felt like `set_data` implies that entire byte buffer should be set, and `set_data_range` means that we might set only portion of it and not necessarily from beginning. 

My original idea was having `set_data_range` and calling this function from `set_data`. I will have a look again to see if I can make it work with a flag.

src/file/reader.rs
sadikovi: Good point. Will change and give variable a more meaningful name.

src/file/reader.rs
sadikovi: Thanks, will change! I did not know there was such function:)

sadikovi:
@sunchao Thanks for the review. I will address your comments and add unittests.

src/encodings/levels.rs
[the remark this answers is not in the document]
sunchao: #[inline] marks the function call as [inline](https://en.wikipedia.org/wiki/Inline_function) so invocation cost can potentially be avoided. This should only be applied to short functions though.

src/encodings/levels.rs
sunchao: Hmm.. on a second thought I think adding a flag may not be that easy. I'm OK with keeping the current solution.

sadikovi:
@sunchao could you review this pull request again? I added tests file/reader.rs and column/reader.rs. 

Thanks!

src/column/reader.rs
```diff
@@ -339,127 +369,170 @@ mod tests {
...
+    let mut tester = ColumnReaderTester::<Int32Type>::new();
+    tester.test_plain_v1(desc.clone(), NUM_PAGES, NUM_LEVELS, 16, ::std::i32::MIN, ::std::i32::MAX);
+    let mut tester = ColumnReaderTester::<Int32Type>::new();
+    tester.test_plain_v2(desc.clone(), NUM_PAGES, NUM_LEVELS, 16, ::std::i32::MIN, ::std::i32::MAX);
   }
 
   #[test]
   fn test_read_plain_int32_uneven() {
-    let mut tester = ColumnReaderTester::<Int32Type>::new();
     let primitive_type = SchemaType::primitive_type_builder("a", PhysicalType::INT32)
```
sunchao: Can we replace this with `get_test_int32_type()` ? I think I missed these in the last commit.

src/column/reader.rs
```diff
@@ -339,127 +369,170 @@ mod tests {
...
   }
 
   #[test]
   fn test_read_plain_int32_multi_page() {
-    let mut tester = ColumnReaderTester::<Int32Type>::new();
     let primitive_type = get_test_int32_type();
-    let descr = ColumnDescriptor::new(
-      Rc::new(primitive_type), None, MAX_DEF_LEVEL, MAX_REP_LEVEL, ColumnPath::new(Vec::new()));
-    tester.test_plain(Rc::new(descr), NUM_PAGES, NUM_LEVELS, 512, ::std::i32::MIN, ::std::i32::MAX);
+    let desc = Rc::new(ColumnDescriptor::new(
```
sunchao: These testing functions are getting a little bloated. All of them are similar except input parameters. Can we define a macro for this? e.g.,

```rust
test_read!(i32, plain_v1, MAX_DEF_LEVEL, MAX_REP_LEVEL, 
  NUM_PAGES, NUM_LEVELS, 512, ::std::i32::MIN, ::std::i32::MAX);
test_read!(i32, plain_v2, MAX_DEF_LEVEL, MAX_REP_LEVEL, 
  NUM_PAGES, NUM_LEVELS, 512, ::std::i32::MIN, ::std::i32::MAX);
test_read!(i32, dict_v1, MAX_DEF_LEVEL, MAX_REP_LEVEL, 
  NUM_PAGES, NUM_LEVELS, 16, 0, 3);
...
```

sadikovi:
@sunchao I addressed your comments - replaced all test functions in `column/reader.rs` with macro calls. Can you review again? Thanks!

sunchao:
Merged. Thanks!

sadikovi:
Thank you for the review and merge!

PR #18, opened 2017-09-19
Fix printer/parser incompatibility

sadikovi:
This PR fixes incompatibility between parser and printer, this is mainly around printing logical types, e.g. fixed byte array and decimal.

Before:
```
message schema {
  OPTIONAL group foo {
    REQUIRED INT32 f1;
    OPTIONAL BYTE_ARRAY f2;
  }
  REPEATED FIXED_LEN_BYTE_ARRAY f3;
}
```

After:
```
message schema {
  OPTIONAL group foo {
    REQUIRED INT32 f1 (INT_32);
    OPTIONAL BYTE_ARRAY f2 (UTF8);
  }
  REPEATED FIXED_LEN_BYTE_ARRAY (12) f3 (INTERVAL);
}
```

Hope this patch fixes printing correctly.

I updated tests to cover parsing of logical types and also added tests for printing -> parsing chain to make sure that we can parse our own printed message type.

I did not add test for printing and parsing column id, because I could not find example of such message type as string with ids.


sadikovi:
@sunchao please review - tried addressing issues with printer <-> parser situation, as you pointed out on one of the reviews.

src/schema/printer.rs
```diff
@@ -217,11 +258,86 @@ mod tests {
...
+              Rc::new(Type::new_primitive_type("b4", Repetition::OPTIONAL,
+                PhysicalType::DOUBLE, LogicalType::NONE, 0, 0, 0, None).unwrap())
+            ], None).unwrap()),
+          ], None).unwrap())
+
+        ], None).unwrap()),
+      ], None
+    ).unwrap();
+
+    let mut s = String::new();
```
sunchao: Since this code is shared, maybe extract it into a separate function?

src/schema/printer.rs
```diff
@@ -140,8 +140,35 @@ impl<'a> Printer<'a> {
...
+          _ => format!("{}", physical_type),
+        };
+        // also print logical type if it is available
+        let logical_type_str = match basic_info.logical_type() {
+          LogicalType::NONE => format!(""),
+          decimal @ LogicalType::DECIMAL => {
+            // for decimal type we should print precision and scale if they are > 0, e.g.
+            // DECIMAL(9, 2) - DECIMAL(9) - DECIMAL
+            let precision_scale = if precision > 0 {
+              if precision > 0 {
```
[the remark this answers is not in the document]
sadikovi: You are right. I am not sure why test passes, will check.

src/schema/printer.rs
sadikovi: Yes, will do. Thanks.

sadikovi:
@sunchao I updated PR to the latest changes in master. Could you please review? Thanks!

sunchao:
Looking good. Merged. Thanks!

sadikovi:
Thanks for merging!

PR #19, opened 2017-09-20
Add builders for primitive & group types

sunchao:
Currently with `new_primitive_type(..)` and `new_group_type(..)`, we always
have to pass all the parameters, even though some of them are usually
default (e.g., precision, scale, length).

This adds builders for constructing `PrimitiveType` and `GroupType`, which
start with a default set of attributes and can append more with
`with_XX(..)` methods.


sunchao:
@sadikovi would you mind reviewing this one?

sunchao:
Thanks. Merged.

PR #20, opened 2017-09-30
DELTA_BINARY_PACKED encoding/decoding

sadikovi:
This PR fixes delta bit pack decoder and adds delta bit pack encoder for `Int32Type` and `Int64Type`.

Encoder implementation is based on description in parquet-format with some initial code from parquet-cpp (which is incomplete in that repository). To reduce code duplication, I created `DeltaBitPackEncoderConversion` trait, which provides some helper methods to compute deltas or convert into unsigned values - both INT_32 and INT_64 implement it.

Decoder is fixed and updated in similar fashion. I created trait `DeltaBitPackDecoderConversion` that has couple of methods for converting value into resulting type (`i32` or `i64`), since most of the code is shared between INT_32 and INT_64.

I also had to update several methods in `bit_utils` to support i64 encoding/decoding; this includes assertions on max `num_bits`, which is reasonable, and MAX_VLQ_BYTE_LEN.

Usage of the delta decoder type is updated for other decoders (`DeltaLengthByteArrayDecoder` and `DeltaByteArrayDecoder`) - checked with parquet-cpp to make sure that change was correct: [len_decoder](https://github.com/apache/parquet-cpp/blob/master/src/parquet/encoding-internal.h#L911) and [prefix_len_decoder](https://github.com/apache/parquet-cpp/blob/master/src/parquet/encoding-internal.h#L965).

Encoding/decoding is tested on local files and unit-tests. I also verified that bytes written by encoder match the same byte array that parquet-mr encoder produces.


sadikovi:
@sunchao I updated delta bit pack encoding/decoding, and only added it for Int32Type. Let me know if PR description is incorrect or missing something - I will change it. Added tests should cover most of cases for encoding/decoding.

Could you review? Thanks!

sunchao:
Thanks @sadikovi for the PR! 

> I tested parquet-mr files (writer version v2), and it looks like INT_32 is decoded with DELTA_BINARY_PACKED, but INT_64 is encoded with PLAIN for the same values.

Not sure about this. From the code it seems parquet-mr uses DELTA_BINARY_PACKED for both int32 and int64. See [here](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java#L96). Also the Parquet encoding format specifies that this encoding is for both int32 and int64; see [here](https://github.com/apache/parquet-format/blob/master/Encodings.md).

So I'm wondering if we should add an impl for i32 instead of replacing the existing impl for i64.

sadikovi:
@sunchao good point. I will update accordingly.

src/encodings/decoding.rs
```diff
@@ -328,7 +328,7 @@ pub struct DeltaBitPackDecoder<T: DataType> {
 
 impl<T: DataType> DeltaBitPackDecoder<T> {
   pub fn new() -> Self {
-    Self { bit_reader: None, num_values: 0, num_mini_blocks: 0,
+    Self { bit_reader: BitReader::from(vec![]), num_values: 0, num_mini_blocks: 0,
```
sadikovi: I replaced Option<BitReader> with just BitReader, because it fixes TODO item and removes assertions on whether or not bit reader is already set. I think that this is equivalent of bit reader being empty (if `set_data` is not called).

src/util/bit_util.rs
```diff
@@ -561,20 +564,6 @@ mod tests {
   }
 
   #[test]
-  #[should_panic(expected = "Num of bytes exceed MAX_VLQ_BYTE_LEN (5)")]
-  fn test_bit_reader_get_vlq_int_overflow() {
```
sadikovi: I removed this test for one reason - could not find example of byte sequence that will panic for MAX_VLQ_BYTE_LEN(10). IMHO, I am not sure if we need this test at all now, since I increased to 10, and it covers all i64 values.

sadikovi:
@sunchao I updated code to support both i32 and i64 encoding/decoding, PR description to reflect new changes and added tests. Could you review again, please? Thanks!

Let me know if I need to change something or add more tests!

src/encodings/encoding.rs
```diff
@@ -331,6 +335,282 @@ default impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+  // it is also okay to encode when no values have been provided
+  fn write_page_header(&mut self) {
+    // write the size of each block
+    self.page_header_writer.put_vlq_int(self.block_size as u64);
+    // write the number of mini blocks
+    self.page_header_writer.put_vlq_int(self.num_mini_blocks as u64);
+    // write the number of all values (including non-encoded first value)
+    self.page_header_writer.put_vlq_int(self.total_values as u64);
+    // write first value
+    self.page_header_writer.put_zigzag_vlq_int(self.first_value);
```
sunchao: wonder if we should silently ignore the return values. Maybe just check the last one?

src/util/bit_util.rs
```diff
@@ -250,13 +250,13 @@ impl BitWriter {
...
-  /// The `num_bits` must not be greater than 32. This is bit packed.
+  /// The `num_bits` must not be greater than 64. This is bit packed.
   ///
   /// Return false if there's not enough room left. True otherwise.
   #[inline]
   pub fn put_value(&mut self, v: u64, num_bits: usize) -> bool {
-    assert!(num_bits <= 32);
-    assert_eq!(v >> num_bits, 0);
+    assert!(num_bits <= 64);
+    assert_eq!(v.checked_shr(num_bits as u32).unwrap_or(0), 0); // covers case v >> 64
```
sunchao: why is `num_bits = 64` not covered by `v >> num_bits`?

src/util/bit_util.rs
```diff
@@ -335,8 +336,9 @@ impl BitWriter {
 
 
 /// Maximum byte length for a VLQ encoded integer
+/// MAX_VLQ_BYTE_LEN = 5 for i32, and MAX_VLQ_BYTE_LEN = 10 for i64
 // TODO: why maximum is 5?
```
sunchao: we can remove the TODO.

src/encodings/decoding.rs
sadikovi: My understanding was that the code will fail anyway, when `set_data()` is not called, because of the reading from empty bit reader. I can add boolean flag to check if `set_data()` was called and assert it in `get()`.

src/encodings/decoding.rs
```diff
@@ -428,25 +407,55 @@ impl Decoder<Int64Type> for DeltaBitPackDecoder<Int64Type> {
...
-      let bit_reader = self.bit_reader.as_mut().unwrap();
-
       // TODO: use SIMD to optimize this?
-      let delta = bit_reader.get_value::<u64>(self.delta_bit_width as usize)
+      let delta = self.bit_reader.get_value::<u64>(self.delta_bit_width as usize)
         .ok_or(eof_err!("Not enough data to decode 'delta'"))?;
-      self.current_value += self.min_delta;
-      self.current_value += delta as i64;
-      buffer[i] = self.current_value;
+      self.current_value = self.current_value.wrapping_add(self.min_delta);
```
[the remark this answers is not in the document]
sadikovi: This is done, because it is okay for values to overflow: [parquet-mr comment](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/delta/DeltaBinaryPackingValuesWriterForLong.java#L89). We should still process current value and/or delta in either decoding or encoding.

For example, this can happen when you have a sequence of i64::MAX and i64::MIN values - the test `encodings::decoding::tests::test_delta_bit_packed_int64_min_max` fails if one changes the code to remove wrapping; however, I think it is perfectly valid to process such sequence of values for decoding.

I am happy to make the change, but then I will have to update tests too.

src/encodings/encoding.rs
```diff
@@ -331,6 +335,282 @@ default impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+  _phantom: PhantomData<T>
+}
+
+impl<T: DataType> DeltaBitPackEncoder<T> {
+  pub fn new() -> Self {
+    let block_size = DEFAULT_BLOCK_SIZE;
+    let num_mini_blocks = DEFAULT_NUM_MINI_BLOCKS;
+    let mini_block_size = block_size / num_mini_blocks;
+    assert!(mini_block_size % 8 == 0);
+    Self::assert_supported_type().unwrap();
```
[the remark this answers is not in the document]
sadikovi: Yes, you are right - will update.

src/encodings/encoding.rs
sadikovi: Do you mean checking boolean result whether or not value was inserted and panic if it was not? If so, then yes, I can add the check and return Err() if `write_page_header()` method fails (fails to write at least one of those values).

Should I add similar check for all `bit_writer.put_***` methods?

src/util/bit_util.rs
sadikovi: Ok, will do. I am still quite surprised that everyone uses MAX_VLQ_BYTE_LEN = 5. It looks like code was copied from Impala codebase and never changed since, but I might be wrong. I am not sure why there is check for MAX_VLQ_BYTE_LEN at all (in Impala and parquet-cpp).

src/util/bit_util.rs
sadikovi: I was puzzled by this as well, potentially because I do not really understand how languages implement right shift. It looks like right shift returns the same result when shift >= width of type - it is okay to shift 63 and 1 for example, but is not okay to shift 64; and it is behaviour for most of the languages out there (found related posts for C++/C, and Java).

Here is Java explanation on this (http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19):
> If the promoted type of the left-hand operand is int, only the five lowest-order bits of the right-hand operand are used as the shift distance. It is as if the right-hand operand were subjected to a bitwise logical AND operator & (§15.22.1) with the mask value 0x1f (0b11111). The shift distance actually used is therefore always in the range 0 to 31, inclusive.

> If the promoted type of the left-hand operand is long, then only the six lowest-order bits of the right-hand operand are used as the shift distance. It is as if the right-hand operand were subjected to a bitwise logical AND operator & (§15.22.1) with the mask value 0x3f (0b111111). The shift distance actually used is therefore always in the range 0 to 63, inclusive. 

I found that Rust has a function `checked_shr` that returns None when such shift is performed, therefore used it to cover that case.

sadikovi:
Thanks for the review. I will address comments shortly.

src/encodings/decoding.rs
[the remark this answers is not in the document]
sunchao: Ah - make sense. I guess it is OK for deltas to contain "overflowed" values as long as `wrapping_add` can restore the original value. Can we add some comments for this code?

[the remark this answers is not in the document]
sadikovi: Will add, thanks!

src/encodings/encoding.rs
sadikovi: I agree, we should update - will fix.

sadikovi:
@sunchao I addressed your comments and updated the code. Can you review again?

sunchao:
Thanks @sadikovi ! I've merged this.

sadikovi:
@sunchao thank you very much for review and merge!

PR #22, opened 2017-10-03
Add check_contains method for Type

sadikovi:
This PR adds `check_contains` method that allows to check if `sub_type` is part of current type. 

Method returns `true` if `sub_type` has the same name and repetition (both have the same value or both have it unset - message type), and for primitive type physical types must match, and for group type all fields in `sub_type` must exist in this type and must match recursively.

This method is designed after parquet-mr `checkContains` that is used to check if projected columns are part of file schema.

Added unit-tests that hopefully cover all usage cases.


sadikovi:
@sunchao I created another PR, hoping that this will be helpful when doing projection on columns. I was planning to use it for naive reader implementation (will create API proposal _issue_ later) in addition to low-level `read_batch` and Arrow.

Let me know if this is useful, and I would appreciate if you could review it. Thanks!

src/schema/types.rs
```diff
@@ -74,6 +74,45 @@ impl Type {
...
+      Type::GroupType { .. } if basic_match && sub_type.is_group() => {
+        // build hashmap of name -> TypePtr
+        let mut field_index = HashMap::new();
+        for field in self.get_fields() {
+          field_index.insert(field.name(), field);
+        }
+
+        let mut found_subtype = true;
+        for field in sub_type.get_fields() {
+          found_subtype = found_subtype && match field_index.get(field.name()) {
```
sunchao: nit: we can quit early if `found_subtype` is false.

sadikovi: Will update, thanks.

src/schema/types.rs
```diff
@@ -74,6 +74,43 @@ impl Type {
...
+      },
+      Type::GroupType { .. } if basic_match && sub_type.is_group() => {
+        // build hashmap of name -> TypePtr
+        let mut field_map = HashMap::new();
+        for field in self.get_fields() {
+          field_map.insert(field.name(), field);
+        }
+
+        for field in sub_type.get_fields() {
+          if !field_map.get(field.name()).map(|tpe| tpe.check_contains(field)).unwrap_or(false) {
```
sadikovi: Updated check, now I am using option to convert into `Option<bool>` and return if false.

src/schema/types.rs
```diff
@@ -74,6 +74,43 @@ impl Type {
...
+            return false;
+          }
+        }
+        true
+      },
+      _ => false,
+    }
+  }
+
+  fn get_physical_type(&self) -> PhysicalType {
```
sadikovi: Is this okay to have this function for `Type` enum?

sunchao: Yes it is OK. Perhaps better to move it after `get_fields()`.

src/schema/types.rs
sunchao: Cool. This is better :)

sadikovi:
@sunchao I addressed comments. Can you review again? Thank!

sunchao:
Thanks @sadikovi ! I've merged this.

sadikovi:
Thanks for merging!

PR #24, opened 2017-10-14
Fix max definition and repetition levels in schema descriptor

sadikovi:
This PR fixes issue when max definition and repetition levels are not resolved correctly. This is observed well on optional struct type fields:
```
OPTIONAL group b {
  OPTIONAL INT32 _1;
  OPTIONAL INT32 _2;
}
```
as well as repeated fields, however, because both repetition and definition levels are incremented, issue might have been masked due to switch def -> rep and rep -> def in the code.

It should be correct now, I tested on several files locally and added unit-test (passes with fix, fails otherwise).


sadikovi:
@sunchao Could you review this PR, please? Thanks!

sunchao:
Merged!

sadikovi:
Thanks for merging! More than happy to help and fix them!

PR #25, opened 2017-10-14
Add BIT_PACKED level encoder/decoder

sadikovi:
This PR adds support for BIT_PACKED level encoder/decoder alongside RLE level encoder/decoder. Bit packed level decoder/encoder is treated as dev/null, and it is used mainly as placeholder for repetition level 0 -> [parquet-mr dev/null encoder](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/DevNullValuesWriter.java). I may be wrong with dev/null observation, but found this behaviour on several files - then repetition level > 0 - RLE, otherwise BIT_PACKED.

This implementation is working (not dummy like in parquet-mr), and allows to write and read levels with bit packing, similar to parquet-cpp.

**Implementation notes**:
- `LevelDecoder` and `LevelEncoder` are now enums for RLE and BIT_PACKED
- `set_data()` method now takes `num_buffered_values` to calculate max length to decode for bit packed decoder, also value is used to keep track of values left
- `set_data_range()` method now takes `num_buffered_values` to keep track of total values left
- `get()` method is implemented as simple for loop, I added TODO to implement `get_batch()`

I added unit-tests for level encoding/decoding. I could not test bit packed level encoding/decoding when max level is greater than 0 on actual files (they are all 0 when it is bit packed).


sadikovi:
@sunchao I would appreciate if you could have a look at this pull request. Thanks!

sadikovi:
@sunchao could you also have a look at this work? I made these changes since some of the files end up with bit packed level encoding. Thanks!

sunchao:
@sadikovi sure. Will take a look soon.

src/encodings/levels.rs
```diff
@@ -15,44 +15,64 @@
...
-/// A encoder for definition/repetition levels. This is a thin
-/// wrapper on `RleEncoder`. Currently only support RLE encoding.
-pub struct LevelEncoder {
-  encoding: Encoding,
-  bit_width: u8,
-  rle_encoder: RleEncoder
+/// A encoder for definition/repetition levels.
+/// Currently only supports RLE and BIT_PACKED (dev/null) encoding.
+pub enum LevelEncoder {
+  LevelEncoderRLE { bit_width: u8, rle_encoder: RleEncoder },
```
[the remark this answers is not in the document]
sadikovi: Yes, sure. I will rename both RLE and Bit packed.

sadikovi:
Should I add more tests? I feel like there are some cases that I have not covered.

sunchao:
> Should I add more tests? I feel like there are some cases that I have not covered.

The current tests look OK for me. The only things I can think of are:
1. in `LevelEncoder.put()`, when `rle_encoder.put()` or `bit_packed_encoder.put_value()` calls return false, therefore `num_decoded` != `buffer.len()`. This doesn't seem to be covered right now.
2. similarly, in `LevelDecoder.get()`, when `buffer.len()` > number of values in the decoder.

BTW: some documents may be needed for the public functions of LevelDecoder/LevelEncoder. But I think we can address that in a separate PR.

Let me know if you have other thoughts :)

sadikovi:
Yes, you are right - I will add test cases. Also agree that we should document functions - will do that.

src/encodings/levels.rs
```diff
@@ -15,47 +15,80 @@
...
+        for v in buffer {
+          if !rle_encoder.put(*v as u64)? { break; }
+          num_encoded += 1;
+        }
+        rle_encoder.flush()?;
+      },
+      LevelEncoder::BIT_PACKED { bit_width, ref mut bit_packed_encoder } => {
+        for v in buffer {
+          if !bit_packed_encoder.put_value(*v as u64, bit_width as usize) {
+            return Err(general_err!("Not enough bytes left"));
```
sadikovi: This is similar to RLE, but in RLE the error is thrown from `rle_encoding.rs`.

src/encodings/levels.rs
```diff
@@ -64,103 +97,274 @@ impl LevelEncoder {
...
   #[inline]
   pub fn get(&mut self, buffer: &mut [i16]) -> Result<usize> {
-    self.rle_decoder.get_batch::<i16>(buffer)
+    match *self {
+      LevelDecoder::RLE { bit_width: _, ref mut rle_decoder } => {
+        rle_decoder.get_batch::<i16>(buffer)
+      },
+      LevelDecoder::BIT_PACKED { bit_width, ref mut num_values, ref mut bit_packed_decoder } => {
+        // check if set data is called
+        assert!(num_values.is_some());
```
sadikovi: Similar to RLE, we check if data is available to read.

src/encodings/levels.rs
```diff
@@ -64,103 +97,274 @@ impl LevelEncoder {
...
+        while i < len {
+          if let Some(value) = bit_packed_decoder.get_value::<i16>(bit_width as usize) {
+            buffer[i] = value;
+            i += 1;
+          } else {
+            break;
+          }
+        }
+        // update current values
+        *num_values = num_values.map(|x| x - i);
```
sadikovi: Used to keep track of current number of values.

src/util/bit_util.rs
```diff
@@ -207,7 +207,7 @@ impl BitWriter {
   #[inline]
   pub fn skip(&mut self, num_bytes: usize) -> Result<usize> {
     self.flush();
-    assert!(self.byte_offset < self.max_bytes);
+    assert!(self.byte_offset <= self.max_bytes);
```
sadikovi: This seems to be the correct check, arrow bit writer has assertion on `less than or equal`.

sadikovi:
@sunchao I added tests and documentation. Would appreciate if you could review them when you have time. 

I left some inline comments on pull request, and overall changes are below:

- Both RLE and BIT_PACKED level decoders have `num_values` field to keep track of total values left in decoder. I found that RLE and bit packed can return more values than there are actually (those additional values are zeros) - added test for it
- Updated several assertion checks
- Made `set_data` for data page v1 and `set_data_range` for data page v2 to take `num_buffered_values` as total number of values that decoder has
- This allows me to add assertion on whether or not `set_data` was called before getting values (initially it was only supported by RLE)

Honestly, these changes become larger and larger, and I am not very confident in my code changes. Could you help me to double check any performance implications or better handling of situations when we need to check if set is called before get, and values increment, etc.? Thanks!

_I will run manual tests on several files to make sure I can still read them with my changes:)._

src/encodings/levels.rs
sunchao: Makes sense. I think `rle_encoder` returns `Err` when the internal buffer is full, which is confusing. I was not consistent in the return value: when the buffer is full, sometimes we return false, while sometimes we return `Err`. In future maybe we should add a new error type such as `BufferOverflow`. But this doesn't have to be addressed in this PR.

src/encodings/levels.rs
```diff
@@ -64,120 +97,352 @@ impl LevelEncoder {
...
+        // update current num_values
+        *num_values = num_values.map(|x| x - values_read);
+        Ok(values_read)
+      },
+      LevelDecoder::BIT_PACKED { bit_width, ref mut num_values, ref mut bit_packed_decoder } => {
+        assert!(num_values.is_some(), "No data set for decoding");
+        // when extracting values from bit reader, it might return more values than left because of
+        // padding to a full byte, we use num_values to track precise number of values.
+        // TODO: Use get_batch() for bit packed decoder
+        let mut i = 0;
```
sunchao: nit: maybe rename `i` to something more meaningful like `values_read`?

src/util/bit_util.rs
sunchao: Yes this looks right.

src/encodings/levels.rs
```diff
@@ -64,120 +97,352 @@ impl LevelEncoder {
...
-/// A encoder for definition/repetition levels. This is a thin
-/// wrapper on `RleDecoder`. Currently it only support RLE encoding.
-pub struct LevelDecoder {
-  encoding: Encoding,
-  bit_width: u8,
-  rle_decoder: RleDecoder
+/// A decoder for definition/repetition levels.
+/// Currently only supports RLE and BIT_PACKED (dev/null) encoding.
+pub enum LevelDecoder {
+  RLE { bit_width: u8, num_values: Option<usize>, rle_decoder: RleDecoder },
```
sunchao: Seems there are some duplication. Maybe we can create an enum just for the internal decoder? i.e.:
```
enum InternalDecoder {
  RLE { decoder: RleDecoder },
  BIT_PACKED { decoder: BitRaeder }
}
```

This also applies to `LevelEncoder`.

src/encodings/levels.rs
```diff
@@ -64,120 +97,352 @@ impl LevelEncoder {
...
+
+  // test when encoded values are larger than encoder's buffer
+  fn test_internal_roundtrip_overflow(encoding: Encoding, levels: &[i16], max_level: i16) {
+    let max_buffer_size = LevelEncoder::max_buffer_size(encoding, max_level, levels.len());
+    let mut encoder = LevelEncoder::new(encoding, max_level, vec![0; max_buffer_size]);
+    let mut found_err = false;
+    // insert a large number of values, so we run out of space
+    for _ in 0..100 {
+      match encoder.put(&levels) {
+        Ok(_) => found_err = false,
```
sunchao: I guess OK branch can just be a no-op.

src/encodings/levels.rs
```diff
@@ -64,120 +97,352 @@ impl LevelEncoder {
...
+
+    let mut decoder = LevelDecoder::new(encoding, max_level);
+    decoder.set_data(levels.len(), ByteBufferPtr::new(encoded_levels));
+    let mut buffer = vec!();
+    let mut total_decoded = 0;
+    let mut safe_stop = levels.len() * 2; // still terminate in case of issues in the code
+    // perform incremental read until all bytes are read
+    while safe_stop > 0 {
+      safe_stop -= 1;
+      let mut tmp = vec![0; 1];
```
sunchao: not sure if `tmp` is necessary - we can just use `&mut buffer[total_decoded..total_decoded+1]`

sadikovi:
@sunchao thanks for the review! I will address comments.

sadikovi:
@sunchao I updated the code, looks better now. I found that it fails when reading one of the examples I have, will try fixing it. Meanwhile, could you have a look, when you have time? Thanks!

sadikovi:
@sunchao I updated function documentation - should be okay now. Also issue was related to repetition levels which is fixed in master, after rebase all my example files worked correctly.

sadikovi:
@sunchao Should we merge this PR?:)

sunchao:
Merged. Thanks @sadikovi for the awesome work!

sadikovi:
@sunchao thank you for reviews and merge!

PR #29, opened 2017-12-07
Add tests

sadikovi:
This PR adds a few simple tests for `file/reader.rs` and `encoding/decoding.rs` and makes minor updates to error messages.


sadikovi:
@sunchao Could you have a look at this PR? Thanks!

src/file/reader.rs
```diff
@@ -515,4 +558,24 @@ mod tests {
     assert!(file.is_ok());
     file.unwrap()
   }
+
+  fn get_temp_file<'a>(file_name: &str, content: &[u8]) -> fs::File {
```
sunchao: Is `<'a>` necessary here?

src/file/reader.rs
```diff
@@ -515,4 +558,24 @@ mod tests {
...
+    // build tmp path to a file in "target/debug/testdata"
+    let mut path_buf = env::current_dir().unwrap();
+    path_buf.push("target");
+    path_buf.push("debug");
+    path_buf.push("testdata");
+    fs::create_dir_all(&path_buf).unwrap();
```
sunchao: Wonder if we need to take care of the case when a path already exist (e.g., calling `get_temp_file` twice in the same fn). Seems in that case it will return `Err`?

src/file/reader.rs
sadikovi: Will remove, thanks.

src/file/reader.rs
sadikovi: I tried calling this method multiple times on the same directory - it is no-op, it is similar to mkdir -p. Plus, I run tests several times and it works! I think it should be okay.

sadikovi:
@sunchao I removed lifetime parameter from both test helper methods. Looks like it works.

src/file/reader.rs
sunchao: Sounds good!

sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks @sunchao for reviewing and merging!

PR #30, opened 2017-12-12
Add CRC32 with SSE support

sunchao:
This adds a CRC32 impl based on the SSE4.2 intrinsics. The hash_util::hash
function now decide at build time which hash to use. To enable the CRC32 hash,
one can pass the flags through RUSTC flag:

  RUSTFLAGS="-C target-feature=+sse4.2" cargo bench

Benchmarking shows that it can give some improvements.

Before:
test dict_i32_1k_10             ... bench:      13,915 ns/iter (+/- 313)
test dict_i32_1k_100            ... bench:      14,208 ns/iter (+/- 385)
test dict_i32_1k_1000           ... bench:      18,119 ns/iter (+/- 9,656)
test dict_i32_1m_10             ... bench:  14,113,301 ns/iter (+/- 272,263)
test dict_i32_1m_100            ... bench:  14,506,331 ns/iter (+/- 315,964)
test dict_i32_1m_1000           ... bench:  19,420,768 ns/iter (+/- 501,641)
test dict_str_1m                ... bench:  12,776,750 ns/iter (+/- 7,132,524)

After:
test dict_i32_1k_10             ... bench:      10,635 ns/iter (+/- 417)
test dict_i32_1k_100            ... bench:      10,778 ns/iter (+/- 133)
test dict_i32_1k_1000           ... bench:      16,894 ns/iter (+/- 223)
test dict_i32_1m_10             ... bench:  11,069,272 ns/iter (+/- 251,322)
test dict_i32_1m_100            ... bench:  11,080,385 ns/iter (+/- 215,262)
test dict_i32_1m_1000           ... bench:  11,445,189 ns/iter (+/- 163,009)
test dict_str_1m                ... bench:  12,963,771 ns/iter (+/- 2,493,590)


sunchao:
@sadikovi Could you take a look at this PR?

src/util/hash_util.rs
```diff
@@ -66,6 +73,42 @@ fn murmur_hash2_64a<T: AsBytes>(data: &T, seed: u64) -> u64 {
...
+/// CRC32 hash implementation using SSE4 instructions. Borrowed from Impala.
+#[cfg(target_feature="sse4.2")]
+pub fn crc32_hash<T: AsBytes>(data: &T, seed: u32) -> u32 {
+  let bytes: &[u8] = data.as_bytes();
+  let u32_num_bytes = ::std::mem::size_of::<u32>();
+  let mut num_bytes = bytes.len();
+  let num_words = num_bytes / u32_num_bytes;
+  num_bytes %= u32_num_bytes;
+
+  let bytes_u32: &[u32] = unsafe {
```
sadikovi: Is it necessary to use `unsafe` block with from_raw_parts? Would it be reasonable to consider any other way of constructing `bytes_u32`?

src/util/hash_util.rs
```diff
@@ -66,6 +73,42 @@ fn murmur_hash2_64a<T: AsBytes>(data: &T, seed: u64) -> u64 {
...
+    offset += 1;
+  }
+
+  offset = num_words * u32_num_bytes;
+  while offset < num_bytes {
+    hash = sse42::mm_crc32_u8(hash, bytes[offset]);
+    offset += 1;
+  }
+
+  // The lower half of the CRC hash has has poor uniformity, so swap the halves
```
sadikovi: nit: double `has` word in the comment.

src/util/hash_util.rs
```diff
@@ -17,11 +17,18 @@
...
 /// Computes hash value for `data`, with a seed value `seed`.
 /// The data type `T` must implement the `AsBytes` trait.
-/// TODO: implement more efficient hash, such as Crc32, using SSE4 instructions.
-pub fn hash<T: AsBytes>(data: &T, seed: u64) -> u64 {
-  murmur_hash2_64a(data, seed)
+pub fn hash<T: AsBytes>(data: &T, seed: u32) -> u32 {
+  #[cfg(target_feature="sse4.2")] {
+    crc32_hash(data, seed)
+  }
+  #[cfg(not(target_feature="sse4.2"))] {
```
sadikovi: Have you considered making another flag as `target_feature="murmur2"` and also have a default value as "murmur2"? IMHO, this would allow to select murmur2 explicitly, rather achieving it by not selecting sse4.2. Just a thought, it looks good either way.

src/encodings/encoding.rs
```diff
@@ -179,7 +179,7 @@ pub struct DictEncoder<T: DataType> {
 
   // Store `hash_table_size` - 1, so that `j & mod_bitmask` is equivalent to
   // `j % hash_table_size`, but uses far fewer CPU cycles.
```
sadikovi: This is interesting! I did not know that.

src/util/hash_util.rs
sunchao: I think `target_features` are hardware features and they are fixed (see [here](https://github.com/rust-lang/rust/issues/29717) for a complete list). The purpose here is that if the target machine support SSE4.2, then during compilation we can enable the faster CRC32C hashing, otherwise we use Murmur hashing. Since there're only two choices, it is not that necessary to define another feature.

src/util/hash_util.rs
sunchao: Either use `unsafe` or we have to move the bits around and construct u32 from 4 u8s. I think it is "safe" to use the `unsafe` block here, since we know the length of the slice.

src/util/hash_util.rs
sunchao: Good catch! Fixed.

sunchao:
> It looks like this is the only feature exposed right now. Is it worth documenting it somehow? We can also > do it later when reviewing docstrings or something like that.

Yes, we can document this in places such as README later.

> Is this possible to specify this option in Cargo.toml? I saw some examples where you could provide > "features" list when using library in other project.

I think you are talking about feature. I'm not very familiar with this but seems it can be used to specify optional dependencies. Also, seems a feature has to be either a package or a group of other features. Here we are using `target_feature`, which is a set of pre-defined hardware features that can be enabled during compilation. I don't think it is necessary to specify it in Cargo.toml.

src/util/hash_util.rs
sadikovi: Yeah, you are right!

sadikovi:
Looks good! Thanks for the explanations.

sunchao:

PR #31, opened 2017-12-12
Use the thrift crate for Cargo

sunchao:
Up to the moment we are using the Thrift github repo for
cargo build. This changes it to use the thrift crate, which
is the idiomatic way for doing this. It is more stable too.


sunchao:
@sadikovi Could you also review this simple PR too?

sunchao:

PR #38, opened 2018-01-30
Add  DELTA_LENGTH_BYTE_ARRAY and DELTA_BYTE_ARRAY encodings

### LINKED ISSUE #33: Add encoder for DELTA_LENGTH_BYTE_ARRAY
We have decoder for this but not encoder.

### LINKED ISSUE #34: Add encoder for DELTA_BYTE_ARRAY
We have decoder for this but not encoder.

sadikovi:
This PR adds `DELTA_LENGTH_BYTE_ARRAY` and `DELTA_BYTE_ARRAY` encodings and fixes some small issues in decodings.

Changes:
- Added `DELTA_LENGTH_BYTE_ARRAY` encoding 
- Added `DELTA_BYTE_ARRAY` encoding
- Added `slice()` method for `ByteArray`
- Reordered list of encodings/decodings to be in the same order in `get_encoder` and `get_decoder`
- Updated `DELTA_BYTE_ARRAY` decoding `get()` method, fixed current_idx update, and modified overall method to handle prefix length 0 (IMHO, it is a bit simpler now)
- Updated test function and generalized for all delta encodings, added tests for delta byte array decoding (tests delta length byte array decoding implicitly)
- Updated generic encoding test to include delta length byte array and delta byte array encoding/decoding

Fixes #33 
Fixes #34


sadikovi:
@sunchao Could you review this PR too? It adds byte array encodings and fixes some issues to make them work. I only tested with unit tests, not with actual files though. Also apologies for changing `get()` method in `DeltaByteArrayDecoder`, it actually only needed couple of changes, but I thought that these changes made method simpler and easier to understand what is going on - happy to revert it to the original code!

Would love to know your ideas and suggestions for the change. Thanks for merging previous patch - it makes this patch work!

src/encodings/decoding.rs
```diff
@@ -653,6 +653,7 @@ impl<> Decoder<ByteArrayType> for DeltaByteArrayDecoder<ByteArrayType> {
       data.start_from(prefix_len_decoder.get_offset()), num_values)?;
     self.suffix_decoder = Some(suffix_decoder);
     self.num_values = num_prefixes;
+    self.current_idx = 0;
```
sadikovi: I am curious if I should be resetting `self.previous_value` here.

sunchao:
Thanks @sadikovi ! and sorry for the delay. I'll take a look at this PR soon :)

sadikovi:
@sunchao all good, no worries:) whenever you have time to review it.

src/encodings/decoding.rs
```diff
@@ -661,33 +662,25 @@ impl<> Decoder<ByteArrayType> for DeltaByteArrayDecoder<ByteArrayType> {
...
-        assert!(self.previous_value.is_some());
-        let previous = self.previous_value.as_ref().unwrap();
-        prefix_slice = Some(Vec::from(previous.as_ref()));
-      }
       // Process suffix
-      // TODO: this is awkward - maybe we should add a non-vectorized API?
-      let mut suffix = vec![ByteArray::new(); 1];
-      let suffix_decoder = self.suffix_decoder.as_mut().unwrap();
-      suffix_decoder.get(&mut suffix[..])?;
+       // TODO: this is awkward - maybe we should add a non-vectorized API?
```
sunchao: nit: indent seems off.

src/encodings/encoding.rs
```diff
@@ -630,6 +636,134 @@ impl DeltaBitPackEncoderConversion<Int64Type> for DeltaBitPackEncoder<Int64Type>
...
+  }
+}
+
+impl Encoder<ByteArrayType> for DeltaLengthByteArrayEncoder<ByteArrayType> {
+  fn put(&mut self, values: &[ByteArray]) -> Result<()> {
+    let lengths: Vec<i32> =
+      values.iter().map(|byte_array| byte_array.len() as i32).collect();
+    self.len_encoder.put(&lengths)?;
+    for byte_array in values {
+      self.data.extend_from_slice(byte_array.data());
```
sunchao: Is it possible to avoid the copying by just storing the pointers?

src/encodings/encoding.rs
```diff
@@ -630,6 +636,134 @@ impl DeltaBitPackEncoderConversion<Int64Type> for DeltaBitPackEncoder<Int64Type>
...
+    }
+    self.prefix_len_encoder.put(&prefix_lengths)?;
+    self.suffix_writer.put(&suffixes)?;
+    Ok(())
+  }
+
+  fn flush_buffer(&mut self) -> Result<ByteBufferPtr> {
+    let mut total_bytes = vec!();
+    // Insert lengths ...
+    let lengths = self.prefix_len_encoder.flush_buffer()?;
```
sunchao: Perhaps we can just use `lengths` instead of having the `total_bytes`?

sadikovi: I could not find a method to merge 2 `ByteBufferPtr`s together, so I did it using additional vector. Should I add a method to merge 2 buffers?

src/encodings/encoding.rs
```diff
@@ -630,6 +636,134 @@ impl DeltaBitPackEncoderConversion<Int64Type> for DeltaBitPackEncoder<Int64Type>
...
+    total_bytes.extend_from_slice(lengths.data());
+    // ... followed by suffixes
+    let suffixes = self.suffix_writer.flush_buffer()?;
+    total_bytes.extend_from_slice(suffixes.data());
+    // Reset previous value
+    self.previous.clear();
```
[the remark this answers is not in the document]
sadikovi: Probably not. I added it, because I thought if you reuse encoder after `flush_buffer` then we should reset previous value. But it looks like we should not. Will remove.

src/encodings/decoding.rs
sadikovi: Added `self.previous_value.clear()` as well as `self.current_idx = 0`. I think some of tests actually cover this.

src/encodings/encoding.rs
sadikovi: I simply do not know how to do it:) I added global vector with data and copied all values into it, because it was simple. Would appreciate if you could suggest the changes that would allow to store pointers.

sadikovi:
@sunchao Would you mind reviewing this PR again? I addressed some of the comments and replied to your inline comments. Would appreciate if you could have a look. Thanks!

src/encodings/encoding.rs
```diff
@@ -530,11 +536,11 @@ default impl<T: DataType> Encoder<T> for DeltaBitPackEncoder<T> {
     let mut buffer = ByteBuffer::new();
     {
       let bit_buffer = self.page_header_writer.buffer();
-      buffer.write(&bit_buffer[0..self.page_header_writer.byte_offset()])?;
+      buffer.write(&bit_buffer[0..self.page_header_writer.bytes_written()])?;
```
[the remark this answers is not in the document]
sunchao: I see. But `self.page_header_writer.flush();` should have taken `bit_offset` into account:
```rust
  #[inline]
  pub fn flush(&mut self) {
    let num_bytes = ceil(self.bit_offset as i64, 8) as usize;
    assert!(self.byte_offset + num_bytes <= self.max_bytes);
    memcpy_value(
      &self.buffered_values, num_bytes,
      &mut self.buffer.data_mut()[self.byte_offset..]);
    self.buffered_values = 0;
    self.bit_offset = 0;
    self.byte_offset += num_bytes;
  }
```

Either way is fine to me though.

src/encodings/encoding.rs
sunchao: I was thinking about changing `self.data` to a `Vec<ByteArray>` (or `Vec<ByteBufferPtr>`) so that `put` does not copy data. This way, we only copy the value data once in `flush_buffer`.

src/encodings/encoding.rs
sunchao: Hmm. You are right. `ByteBufferPtr` is immutable so we cannot concatenate it. I think we can leave it for now. Could you add a TODO here so in future we can revisit this if it matters?

src/encodings/encoding.rs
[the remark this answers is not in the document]
sadikovi: Yeah, you are right. I will revert those changes - they actually do not affect anything.

src/encodings/encoding.rs
sadikovi: Will try this approach, thanks.

src/encodings/encoding.rs
sadikovi: Will do.

sadikovi:
@sunchao Thank you for the review. I updated the code according to your changes. 
- Reverted `bytes_written`, it did not affect anything + there is no need to mix changes in one PR
- Added TODO item, it looks like it is applicable for both delta length byte array and delta byte array encodings
- Now I clone byte arrays in `put` method in delta length byte array encoding and only copy data when `flush_buffer` is called.

Could you review again, when you have time? Thanks. 
Feel free to comment on any changes, because it helps me to avoid "Java-like" coding, and start thinking on how to avoid extensive data copy.

src/encodings/encoding.rs
```diff
@@ -630,6 +636,136 @@ impl DeltaBitPackEncoderConversion<Int64Type> for DeltaBitPackEncoder<Int64Type>
...
+  }
+}
+
+impl Encoder<ByteArrayType> for DeltaLengthByteArrayEncoder<ByteArrayType> {
+  fn put(&mut self, values: &[ByteArray]) -> Result<()> {
+    let lengths: Vec<i32> =
+      values.iter().map(|byte_array| byte_array.len() as i32).collect();
+    self.len_encoder.put(&lengths)?;
+    for byte_array in values {
+      self.data.push(byte_array.clone());
```
sadikovi: I assume that clone for byte array, triggers clone for byte buffer ptr, which only clones Rc, not the actual data.

sunchao: @sadikovi yes that's right. It will not clone the actual data.

sadikovi:
@sunchao Thanks for merging!

PR #43, opened 2018-02-08
Fix line width in files

sadikovi:
This PR is a minor update, fixes remaining line width in files (<= 90). I assume that this rule does not apply to imports (only line 23 in printer.rs).

Also updated code to use `get_test_column_desc_ptr()` instead of custom code in `test_get_decoder`.


sadikovi:
This is super minor lazy patch of line widths. I thought I might as well do it - does not take long.
@sunchao Could you review, please?

I am not sure if there is a way to find out such files with cargo or IDE - I just used this :):
```python
import os

def check_line_width(f, lines):
    for i, line in enumerate(lines):
        if len(line) > 90:
            print "%s: L%s line width %s > 90" % (f, i+1, len(line))

def check_repo(path, ext=".rs"):
    for root, dirs, files in os.walk(path):
        for f in files:
            if f.endswith(ext) and f != "parquet.rs":
                with open(os.path.join(root, f), 'r') as h:
                    check_line_width(f, [line.strip('\n') for line in h])

check_repo("/path/to/parquet-rs")
```

sadikovi:
Thanks! I never used `rustfmt` - sounds like something we could use.

PR #44, opened 2018-02-19
Add rustfmt and update travis to include format

sadikovi:
This PR adds `rustfmt`  to the project, so we can check style for Rust code easier. Following has been done:
- Added `rustfmt.toml` describing current format, I commented out version, so we always use the latest.
- Updated `.travis.yml` to include a step of running `cargo fmt` to report any inconsistencies.
- Updated `Makefile` and `README`.
- Formatted files by running `make fmt` or `cargo fmt` on current files.

I tried preserving existing style as much as I could, but I still do not know how to fix some things, e.g. commas on the last entries in `match` - but I do not think it is a big deal.


.travis.yml
```diff
@@ -19,9 +19,11 @@ before_script:
   - cd $TRAVIS_BUILD_DIR/src/parquet_thrift && thrift --gen rs parquet.thrift
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
   - cargo install cargo-travis --force && export PATH=$HOME/.cargo/bin:$PATH
+  - rustup component add rustfmt-preview
 
 script:
   - cd $TRAVIS_BUILD_DIR
+  - travis-cargo fmt -- --write-mode=diff
```
sunchao: What happens if the PR fails the style check? will the travis job fail?

rustfmt.toml
```diff
@@ -0,0 +1,55 @@
+max_width = 90
```
sunchao: Are these standard configurations? what are the customizations we are making?

.travis.yml
sadikovi: Yes, the travis job should fail, this command returns non-zero exit code when there are failures. I tested locally, will do the same on travis.

rustfmt.toml
sadikovi: No, I had to change some of them. Changes are below (`<` - our file, `>` - default style):

```
1c1
< max_width = 90
---
> max_width = 100
```

```
3c3
< tab_spaces = 2
---
> tab_spaces = 4
```

```
8,9c8,9
< wrap_comments = true
< comment_width = 90
---
> wrap_comments = false
> comment_width = 80
```

```
19,20c19,20
< reorder_imports = true
< reorder_imports_in_group = true
---
> reorder_imports = false
> reorder_imports_in_group = false
```

```
49c49
< # required_version = "0.3.8"
---
> required_version = "0.3.8"
```

```
55c55
< error_on_unformatted = true
---
> error_on_unformatted = false
```

sunchao:
@sadikovi is this ready to merge? seems it failed the style check?

sadikovi:
Hi @sunchao, it looks like `travis-cargo` complains about passing arguments. I suggest we keep it open for now, if possible. Thanks. 

I just need to figure out how make it work with travis:)

sadikovi:
I am going to close it for now.

PR #46, opened 2018-02-20
Fix default impl issue

### LINKED ISSUE #45: Build fails with Rust 1.25-nightly
Looks like build fails on the latest nightly version:
```
nightly-x86_64-unknown-linux-gnu (default)
rustc 1.25.0-nightly (27a046e93 2018-02-18)
```

Get following errors:
```rust
error[E0599]: no method named `set_data` found for type `encodings::decoding::PlainDecoder<T>` in the current scope
   --> src/column/reader.rs:443:18
    |
443 |       dictionary.set_data(page.buffer().clone(), num_values as usize)?;
    |                  ^^^^^^^^
    | 
   ::: src/encodings/decoding.rs:83:1
    |
83  | pub struct PlainDecoder<T: DataType> {
    | ------------------------------------ method `set_data` not found for this
    |
    = note: the method `set_data` exists but the following trait bounds were not satisfied:
            `encodings::decoding::PlainDecoder<T> : encodings::decoding::Decoder<_>`
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following trait defines an item `set_data`, perhaps you need to implement it:
            candidate #1: `encodings::decoding::Decoder`

error[E0275]: overflow evaluating the requirement `encodings::decoding::PlainDecoder<T>: encodings::decoding::Decoder<T>`
   --> src/column/reader.rs:446:24
    |
446 |       decoder.set_dict(Box::new(dictionary))?;
    |                        ^^^^^^^^^^^^^^^^^^^^
    |
    = note: required because of the requirements on the impl of `encodings::decoding::Decoder<T>` for `encodings::decoding::PlainDecoder<T>`
    = note: required for the cast to the object type `encodings::decoding::Decoder<T>`

error: aborting due to 2 previous errors

error: Could not compile `parquet`.
```

sunchao:
This fixes #45 , an issue caused by the usage of `default impl` on struct, introduced by https://github.com/rust-lang/rust/pull/45404. With the change, `default impl Foo for Bar` doesn't mean `Bar` can be used as trait `Foo`.


sunchao:
@sadikovi Can you review this PR? thanks.

sadikovi:
Looks great! Thanks for fixing this!

sunchao:
Merged. Thanks for the review @sadikovi !

PR #55, opened 2018-03-09
Update set_current_page_encoding() method to get decoder

sadikovi:
This PR replaces `match` statement with just a function `get_decoder()` because it already tests what decoding are available and supported. Thus, we are consistent with across the code base.


sadikovi:
@sunchao I removed TODO by simply using `get_decoder()` method. I do not think we need any additional tests for it:).

sadikovi:
Thanks for merging!

PR #60, opened 2018-03-17
Add support for RLE value encoding/decoding

sadikovi:
This PR adds support for RLE value encoding/decoding. This type of encoding is used primarily for Data Page v2, and allows to read data page v2 pages with boolean values (tested manually on `read-support` branch).

Only boolean values are supported by this encoder/decoder (similar to parquet-mr).

Current implementation of RLE value decoder is based on RLE decoder (essentially a wrapper on it). RLE value encoder is a wrapper on RLE encoder.

Also made some minor style updates to the code, and renamed `rle_encoding.rs` into `rle.rs`. Let me know if this is acceptable. If not, then I will revert it!

Added unittests and also tested manually on `data/test_datapage_v2.snappy.parquet`.


sadikovi:
@sunchao Could you review this PR when you have time? Thanks!

This adds `RleValueEncoder` and `RleValueDecoder` primarily to read data page v2 files with boolean columns.

src/encodings/encoding.rs
```diff
@@ -351,6 +354,68 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+    // Flush all encoder buffers and raw values
+    rle_encoder.flush()?;
+    self.raw_values_buffer.clear();
+
+    // Copy data into bytebuffer
+    let mut encoded_data = Vec::new();
+    let len = (rle_encoder.len() as i32).to_le();
+    let len_bytes = len.as_bytes();
+    encoded_data.extend_from_slice(len_bytes);
+    encoded_data.extend_from_slice(&rle_encoder.consume()[..]);
```
sunchao: maybe we can call `buffer()` instead of `consume()`? this also allows use to reuse the same `rle_encoder`, and avoid repeated allocation of its internal buffer.

src/encodings/encoding.rs
```diff
@@ -351,6 +354,68 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
 }
 
+// ----------------------------------------------------------------------
+// RLE encoding
+
+const DEFAULT_RLE_BUFFER_LEN: usize = 1024;
+
+/// RLE encoding for values.
+/// Currently is used only for data pages v2 and supports boolean types.
+pub struct RleValueEncoder<T: DataType> {
```
sunchao: Seems this may also encode the input data with bit-packed encoding? if so, we may want to clarify in the comments.

src/encodings/encoding.rs
```diff
@@ -351,6 +354,68 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+  #[inline]
+  default fn flush_buffer(&mut self) -> Result<ByteBufferPtr> {
+    panic!("RleValueEncoder only supports BoolType");
+  }
+}
+
+impl Encoder<BoolType> for RleValueEncoder<BoolType> {
+  #[inline]
+  fn flush_buffer(&mut self) -> Result<ByteBufferPtr> {
+    let bit_width: u8 = 1;
```
sunchao: nit: `bit_width` is not necessary - just use 1 for the following line.

src/encodings/encoding.rs
```diff
@@ -351,6 +354,68 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+    let mut rle_encoder = RleEncoder::new(bit_width, DEFAULT_RLE_BUFFER_LEN);
+    // Encode data
+    for value in &self.raw_values_buffer {
+      if !rle_encoder.put(*value as u64)? {
```
sunchao: if `put` returns false, it may mean `rle_encoder` buffer full (although I just found out we didn't implement the logic in `RleEncoder`..). We should handle that case (add as a TODO is fine).

src/encodings/decoding.rs
```diff
@@ -303,6 +311,70 @@ impl<T: DataType> Decoder<T> for DictDecoder<T> {
...
+  _phantom: PhantomData<T>
+}
+
+impl<T: DataType> RleValueDecoder<T> {
+  pub fn new() -> Self {
+    Self { values_left: 0, decoder: None, _phantom: PhantomData }
+  }
+
+  #[inline]
+  fn internal_set(&mut self, data: ByteBufferPtr, num_values: usize) -> Result<()> {
```
sunchao: nit: can we rename this to `set_data_internal`? it has the same signature as `set_data`.

sunchao:
Thanks @sadikovi . Is this covered in the Parquet format standard? I found it only implemented in mr but not cpp. The `rle_encoding.rs` renaming is fine.

src/encodings/decoding.rs
```diff
@@ -702,12 +774,13 @@ mod tests {
   use schema::types::{ColumnDescriptor, ColumnPath, Type as Tpe};
 
   #[test]
-  fn test_get_decoder_int32() {
+  fn test_get_decoder_method() {
```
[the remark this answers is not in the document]
sadikovi: It conflicts with another method `test_get_decoder`. I will rename it into `test_get_decoders` instead, if that is okay.

src/encodings/encoding.rs
sadikovi: I will fix it. What should be the expected behaviour?

src/encodings/encoding.rs
sadikovi: I added the note that it is a hybrid encoding (bit packed + rle).

src/encodings/encoding.rs
sadikovi: Yes, you are right. I will change it.

sadikovi:
Hi @sunchao. Thanks for the review!

Parquet format just mentions that this encoding supports "boolean values in data pages, as an alternative to PLAIN encoding". I only observed this on Data Page v2 (`data/test_datapage_v2.snappy.parquet`).

I tried addressing all your comments. Let me know if I understood you correctly. I updated `RleValueEncoder` to use `rle_encoder` directly instead of some buffer of raw values - had to add `flush_buffer` method for `RleEncoder`. 

Also I updated the for loop in `levels.rs` and `encoding.rs`, where we put values into RLE encoder. Now if encoder is full we return an error instead of just breaking from the loop.

src/encodings/rle.rs
```diff
@@ -189,6 +189,14 @@ impl RleEncoder {
     self.bit_writer.consume()
   }
 
+  /// Borrow equivalent of the `consume` method.
+  /// Call `clear()` after invoking this method.
+  #[inline]
+  pub fn flush_buffer(&mut self) -> &[u8] {
+    self.bit_writer.flush();
```
sunchao: can we also add a `flush_buffer()` for `bit_writer`, which flushes the buffer and return its borrowed reference? this pattern occurs a few times in `encoding.rs`.

Also, maybe this `flush_buffer` should call `flush` first? it's a little confusing that we need to call `flush` and then call `flush_buffer`.

sadikovi: Yes, I will add the method and update occurrences.

src/encodings/rle.rs
```diff
@@ -189,6 +189,13 @@ impl RleEncoder {
     self.bit_writer.consume()
   }
 
+  /// Borrow equivalent of the `consume` method.
+  /// Call `clear()` after invoking this method.
+  #[inline]
+  pub fn flush_buffer(&mut self) -> &[u8] {
+    self.bit_writer.flush_buffer()
```
sunchao: Sorry, I was thinking that this can also call `flush` first:

```rust
pub fn flush_buffer(&mut self) -> Result<&[u8]> {
  self.flush()?
  self.bit_writer.flush_buffer()
}
```
So that it matches the behavior of other `flush_buffer()`s (otherwise, have to call `flush()` before `flush_buffer()` is a little confusing.) However this makes `flush_buffer()` a little hard to change though because of lifetime error. I made it work in this way:
```rust
  #[inline]
  fn flush_buffer(&mut self) -> Result<ByteBufferPtr> {
    assert!(self.encoder.is_some(), "RLE value encoder is not initialized");
    let rle_encoder = self.encoder.as_mut().unwrap();

    // Flush all encoder buffers and raw values
    let encoded_data = {
      let buf = rle_encoder.flush_buffer()?;

      let len = (buf.len() as i32).to_le();
      let len_bytes = len.as_bytes();
      let mut encoded_data = Vec::new();
      encoded_data.extend_from_slice(len_bytes);
      encoded_data.extend_from_slice(buf);
      encoded_data
    };

    // Reset rle encoder for the next batch
    rle_encoder.clear();

    Ok(ByteBufferPtr::new(encoded_data))
  }
```

Also, `test_rle_consume_flush_buffer` needs to call `flush()` on `encoder2` before `consume()`.

Let me know what you think.

sadikovi: All good. Looks interesting. I will try it soon, thanks.

sadikovi: Do we really need to make this change? I was thinking that this might force us to change `consume()` method to behave the same way. Maybe we should simply rename `flush_buffer` to something like `get_buffer`? Or should I proceed with this change?

sunchao: Hmm. Good point. IMO we should change `consume()` too - just like `bit_util::flush_buffer()`, `bit_util::consume()` also flushes before consuming it. So, the `rle::consume()` and `rle::flush_buffer()` should just directly call the corresponding ones in `bit_util`. Let me know what you think.

sadikovi:
@sunchao I updated the code for consume vs flush_buffer. Can you review this commit: fbb883b1a575ef87a4ff54f2dffa274fc36f0c38? I made `consume` to be the same as `flush_buffer` - they both call `flush` before returning data in both `bit_util` and `rle`. Thanks!

sunchao:
I've merged this change. Thank you so much @sadikovi !

sadikovi:
Thanks for merging!

PR #62, opened 2018-03-19
Remove lifetime restriction in file reader

sunchao:
Currently there's borrowing relation between FileReader and
RowGroupReader, due to the sharing of metadata. This restriction is not
necessary as row group reader should be allowed to out-live file reader.
This removes it by replacing the metadata with RC (reference-counting)
values.


src/file/reader.rs
```diff
@@ -21,7 +21,11 @@ use std::rc::Rc;
 
 use basic::{Type, Compression, Encoding};
 use errors::{Result, ParquetError};
-use file::metadata::{RowGroupMetaData, FileMetaData, ParquetMetaData};
+use file::metadata::{
```
sadikovi: How do you feel about using `use file::metadata::*` here?

sunchao: Yes. Makes sense. I'll replace with star-import.

sunchao:
I think this will solve the lifetime issue for the global iterator (since there will no longer be lifetime parameter for `RowGroupReader`), but haven't tried it yet...

sunchao:

sadikovi:
All good! Great to see this update.

PR #64, opened 2018-03-22
Adjust batch size for each iteration in column reader

sadikovi:
This PR fixes the issue, when during call of `read_batch`, we end up buffering a new page. Page buffering (`has_next()` method) results in setting `self.num_decoded_values = 0`, which results in selecting `iter_batch_size = batch_size`. But here is a catch: sometimes actual length of the slices (either values or levels - affects both) is less than `levels_read + iter_batch_size` (or alternatively `values_read + iter_batch_size`). 

Current approach is adjusting batch size for each iteration, so we can take remaining length of slices into account.

This bug is quite difficult to reproduce, I found it on one of my large files, and managed to add test case that reproduces the problem - if you try this test on current master branch it would panic, and passes with this patch!

> P.S. Actually it also fails on previous version of `read_batch`, with assertion error, so this fix should help us.


sadikovi:
@sunchao Could you review this PR, please? Thanks!

This is the issue I found when testing my `read-support` branch.

src/column/reader.rs
```diff
@@ -164,10 +164,25 @@ impl<'a, T: DataType> ColumnReaderImpl<'a, T> where T: 'static {
       }
 
       // Batch size for the current iteration
-      let iter_batch_size = ::std::cmp::min(
-        batch_size,
-        (self.num_buffered_values - self.num_decoded_values) as usize
-      );
+      let iter_batch_size = {
+        // Compute approximate value based on values decoded so far
+        let mut adjusted_size = ::std::cmp::min(
```
sunchao: nit: we have quite a lot `::std::cmp::min` and similar now - perhaps we can import it can replace them with `min` - makes it easier for reading.

src/column/reader.rs
```diff
@@ -164,10 +164,25 @@ impl<'a, T: DataType> ColumnReaderImpl<'a, T> where T: 'static {
...
+          batch_size,
+          (self.num_buffered_values - self.num_decoded_values) as usize
+        );
+
+        // Adjust batch size by taking into account how much space is left in values
+        // slice or levels slices (if available)
+        adjusted_size = ::std::cmp::min(adjusted_size, values.len() - values_read);
```
sunchao: seems this makes line 150 - 157 redundant - perhaps we can remove those (but might need to change `iter_batch_size` to `batch_size` or something similar).

sadikovi: Those lines are still required, they adjust batch size as the smallest value between `batch_size`, `values`, `def_levels`, and `rep_levels`. Without it we might not exist the while loop if `batch_size` > `values_read` or `levels_read`, for example.

src/column/reader.rs
sadikovi: Done.

src/column/reader.rs
sunchao: You are right. That's why I think you need to change `let iter_batch_size = ...` to `batch_size = ...`, and make the parameter `batch_size` mutable.

sadikovi: Sorry, I am not trying to avoid changing the code. I am more than happy to update it!

I was considering this initially, but then I thought that it would affect condition in the while loop, plus, since we decrease batch size with a series of `min(...)`, we would be inserting very small number of elements, even though we could insert full batch size. For example,

Page 1 has 3 values left, batch size is 30 (and we can buffer all 30 values), we would update batch size to be 3, as min(3, 30), read all those values, and then buffer page 2, which will have 50 values, but we would keep inserting by 3, rather than selecting 27, because of min(3, 50).

Let me know if my comment above does not make sense, or it is incorrect in terms of workflow of `read_batch`.

sunchao: Hmm you are right. Sorry for the wrong suggestion. Let's leave it then - we can think about better ways to remove this duplication in future.

sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks for merging @sunchao!

PR #65, opened 2018-03-23
Remove lifetime from column reader

sadikovi:
This PR removes `'a` lifetime from `ColumnReader` and `ColumnReaderImpl`. It looks like we may not need it, and it also helps solving problem of returning a global iterator of records.


sadikovi:
@sunchao Could you review this PR when you have time? As discussed, here we remove `'a` from column reader.

sadikovi:
Thanks!

PR #67, opened 2018-03-24
Add build.rs to generate parquet.rs

### LINKED ISSUE #32: Investigate on publish the project in Cargo crate
At some point we should investigate on the possibility to publish parquet-rs as a crate so that it can be more easily used by other projects. So far I think one main issue is that `parquet.rs` needs to be generated before compile time and there is no easy way to do this (something similar was done for protobuf rust which we may borrow). Otherwise, we have to put the generated `parquet.rs` in the repo, which is not very desirable.

sadikovi:
This PR makes it easy to build with just `cargo` by adding `build.rs` script, which is triggered whenever script itself is changed or source files have been changed. Generation is very quick, so there is no performance impact from doing it this way. When this project is used as a dependency, when generation is triggered only when entire rebuild is required, e.g. after `cargo clean`.

If Thrift is not installed on a machine, script will panic and print the error message saying what is required.

This change will allow to publish `parquet-rs` on crates.io, so people could start using it!

I also updated README by removing manual installation instructions and just pointing to the relevant sources, because they contain more information, including platform specific installs; and removed Makefile, since we could use `cargo` for everything.

Fixes #32


build.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::fs;
+use std::process::Command;
+
+fn main() {
+  let mut curr_dir = env::current_dir().expect("Get project directory");
+  curr_dir.push("src");
+  curr_dir.push("parquet_thrift");
+
+  // Check if thrift is installed
+  if let Err(error) = run(Command::new("which").arg("thrift")) {
```
sunchao: We also need to check Thrift version here. Maybe we can do `thrift version` and check the major/minor version number.

README.md
```diff
@@ -6,57 +6,29 @@
...
+See [Working with nightly Rust](https://github.com/rust-lang-nursery/rustup.rs/blob/master/README.md#working-with-nightly-rust)
+to install nightly toolchain and set it as default. Follow instructions on [Apache Thrift](https://thrift.apache.org)
+website to install the required version of Thrift (it may be necessary to build from source).
 
-Building project requires Thrift 0.11.0 or higher, which comes with the Rust
-support. To build from source:
-```shell
-git clone --depth=1 -b 0.11.0 https://github.com/apache/thrift
-cd thrift
+For more information on setup and build refer to [.travis.yml](./.travis.yml) file
```
sunchao: nit: can we embed the line number in the file link, such as ./.travis.yml#L15?

build.rs
sunchao: We may also need to check version number, such that the major version should be greater than 1 or that the minor version should be greater than 10. You can also just add a TODO here.

sadikovi: Yes, will do. Thanks.

sadikovi:
@sunchao I addressed your comments. Could you have a look again? Thanks!

sunchao:
Merged! Thanks @sadikovi !

PR #70, opened 2018-03-31
Format the source code

sadikovi:
This PR formats the source code based on `rustfmt` rules, files were inspected manually, however.
Most of the formatting is preserved.


sadikovi:
This relates to #44.

sadikovi:
@sunchao Could you review this PR when you have time? I use `cargo fmt -- --write-mode=diff` now, instead of `travis-cargo`. Thanks!

benches/decoding.rs
```diff
@@ -16,124 +16,305 @@
...
+        .get(&mut values[..])
+        .expect("get() should be OK") < batch_size
+      {
+        break;
       }
     }
   })
 }
 
+plain!(
```
sunchao: do you know why it changed to this format? it seems unnecessary.

src/column/reader.rs
```diff
@@ -367,36 +405,46 @@ impl<T: DataType> ColumnReaderImpl<T> where T: 'static {
...
+      .as_mut()
       .expect("def_level_decoder be set");
     level_decoder.get(buffer)
   }
 
   #[inline]
   fn read_values(&mut self, buffer: &mut [T::T]) -> Result<usize> {
-    let encoding = self.current_encoding.expect("current_encoding should be set");
-    let current_decoder = self.decoders
+    let encoding = self
```
sunchao: this also seems unnecessary as it already fits within 90 chars.

src/util/test_common.rs
```diff
@@ -74,20 +73,18 @@ pub fn random_bools(n: usize) -> Vec<bool> {
...
   result
 }
 
-pub fn random_numbers_range<T>(
-  n: usize,
-  low: T,
-  high: T,
-  result: &mut Vec<T>
-) where T: PartialOrd + SampleRange + Copy {
+pub fn random_numbers_range<T>(n: usize, low: T, high: T, result: &mut Vec<T>)
```
sunchao: I like the old format better (even though the parameters could be put in a single line).

sadikovi:
@sunchao I agree with your comments - I actually had to correct the formatted code in a couple of places as well. Let me review the changes again and see if I can tweak the configuration file to make rust format make changes less aggressively.

src/util/test_common.rs
sadikovi: Unfortunately, I do not know how to enforce formatting of function arguments, apart from using `#[cfg_attr(rustfmt, rustfmt_skip)]` for each item.

sadikovi:
@sunchao I updated the `rustfmt.toml` to address most of your comments. I also inspected changes manually, and they look okay for the most part (~ 96%).

Could you have a look again and advice on the next steps, e.g. whether or not we even need this? Thanks!

sunchao:
Hi @sadikovi : sorry for the late reply! Is there a way to just enforce max line width + tab width (also maybe the reordering on imports)? my concern is that it currently enforces too many rules and limits flexibility (e.g., sometimes it may be more concise to put everything in a single line but this will break it up into multiple lines for all cases).

sadikovi:
You are right 100%, I thought about something along those lines. I will try updating the config to check line width, indentation and maybe couple of other basic rules, and disable the rest.

Will update today, thanks.

sadikovi:
@sunchao I could not find the option that would allow to apply only a subset of rules, without applying the rest. It looks like `rustfmt` always applies all rules, but checks for values in provided configuration file, if options do not exist in the file, `rustfmt` uses default values for those options.

Possible solutions:
1. Keep the current rustfmt configuration and changes
2. Close PR and do not apply rustfmt
3. Create a PR with manual formatting based on rustfmt (only imports, indentations and couple of other items)

Let me know which way you would prefer to go with. I personally think that choosing option 3 might be a good way to start.

sunchao:
@sadikovi : I'm also in favor of option 3) for now. We can keep the issue open and come back later if there's better solution. I'm also going to look into why `rustfmt` always apply all rules.

Thanks for working on this!

sadikovi:
@sunchao could you review this PR again? Thanks!
I updated the code manually without any rust format, and this does not include any changes to the build.

sadikovi:
Thanks @sunchao! Would you mind merging this PR, if it is okay? If there is anything you want me to change, let me know - I will update the code. Thanks.

sunchao:
Merged! Thanks!, and sorry for the delay @sadikovi .

sadikovi:
Thank you very much @sunchao for merging!

PR #81, opened 2018-04-10
Update documentation

### LINKED ISSUE #58: Add more comments for test set up for column reader, encoding/decoding
Some of the test setup, such as the one in `column/reader.rs`, is quite complicated and hard to understand. We should add more high-level comments for future reference.

### LINKED ISSUE #77: Suggestion: Add examples
With a new Rust project I usually go look at the examples first. I think it would be helpful to have some simple examples for reading parquet files. Perhaps one column-oriented and one row-oriented?

### LINKED ISSUE #78: Update crate documentation
Currently we do not publish documentation when releasing a new version of crate. It would be very useful for people to get started, if we had one.

Items to cover:
- Compile documentation for crate, so it can be published, e.g. module documentation, structs, methods, etc.
- Add some references in README.
- Add badge to point to the latest docs.
- Fix docs.rs build somehow (?), see comments below.

sadikovi:
This PR updates crate documentation and makes some very minor changes, either in code or comments.

Following is done:
- Updated docs, so now we can generate more or less usable docs with `cargo doc`. I learned that it is good to run `cargo doc --no-deps`, to generate docs for this crate only when developing.
- Added small example in README + usage + link to documentation.
- Added comment for `column/reader.rs` tests to give an overview on how setup works.

I would encourage to view changes as generated documentation as well as code. You can check out this branch and run `cargo doc --no-deps --open` to open documentation in your browser.

Closes #58 
Closes #77 
Closes #78


sadikovi:
@sunchao Could you review this PR? The goal of this work is having documentation/some starting point for people who want to read Parquet files.

## Publishing docs

I added badge in README that points to docs.rs for parquet (it is broken currently). I thought about this and I think that publishing on docs.rs is arguably the best way to do it right now, which is what you also mentioned on the issue #78 . 

My main points are:
- No setup required, it is sort of generated automatically as we release.
- Most of the people publish on docs.rs.

What stops us from doing this currently is generating `parquet.rs` file. Though I understand why you would want to have it generated (this is how it works right now), I would suggest that we actually commit the `parquet.rs` as part of the repository alongside `parquet.thrift`.

When we need to update format, we will create a special PR that updates `parquet.rs` only and potentially fixes code to work with updated format. We will document or have a script that would generate `parquet.thrift` -> `parquet.rs` and will run manually when we decide to update the format.

Reasons for committing `parquet.rs` into repository:
- Fixes documentation build.
- Simplifies travis setup, and speeds up build time (it is not really a big deal, tests run reasonably fast anyway).
- `parquet.rs` does not change often, actually only changes when we change format.
- Rarely thrift repo may be broken, so in this case we would not notice the impact of those breaking changes.
- Assumption that `parquet.rs` is generated identically regardless of OS.
- Helps other people to import the crate, without requirement of setting up thrift (and we would only use it to generate single file), all they would need is just adding `parquet` to their Cargo.toml.

Let me know your thoughts, as you are more familiar with overall setup and `parquet.thrift` generation. If so, I will be more than happy to open other pull requests to update the code.

src/basic.rs
```diff
@@ -40,56 +49,174 @@ pub enum Type {
...
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Encoding {
+  /// Default byte encoding.
+  /// - BOOLEAN - 1 bit per value, 0 is false; 1 is true.
+  /// - INT32 - 4 bytes per value, stored as little-endian.
+  /// - INT64 - 8 bytes per value, stored as little-endian.
+  /// - FLOAT - 4 bytes per value, stored as little-endian.
+  /// - DOUBLE - 8 bytes per value, stored as little-endian.
+  /// - BYTE_ARRAY - 4 byte length stored as little endian, followed by bytes.
+  /// FIXED_LEN_BYTE_ARRAY - just the bytes are stored.
```
sunchao: nit: missed `-` at the beginning.

src/lib.rs
```diff
@@ -37,19 +131,17 @@ extern crate flate2;
...
 
-// TODO: don't expose everything!
 #[macro_use]
 pub mod errors;
 pub mod basic;
 pub mod data_type;
 mod parquet_thrift;
 
 #[macro_use]
-pub mod util;
```
sunchao: This will break `cargo bench`. We need to find a way to expose only the necessary APIs.

src/schema/types.rs
```diff
@@ -669,8 +738,8 @@ fn build_tree(
   }
 }
 
-/// Conversion from Thrift equivalents
-
+/// Conversion from Thrift equivalents.
+/// Should be considered private.
```
sunchao: Why `pub` can't be removed? add a TODO?

sadikovi:
I just need to document `encoding` and `decoding` modules, and review again.

sadikovi:
@sunchao I think it is ready for review now. Would you mind having a look again? Thanks!

README.md
```diff
@@ -4,9 +4,40 @@
...
+let file = File::open(&Path::new("/path/to/file")).unwrap();
+let reader = SerializedFileReader::new(file).unwrap();
+let mut iter = reader.get_row_iter(None).unwrap();
+while let Some(record) = iter.next() {
+  println!("{}", record);
+}
+```
+See crate documentation on available API.
+
+## Versions
```
sunchao: instead of `Versions`, maybe `Supported Parquet Version`?
Also, the version we support right now is 2.4.0

src/schema/types.rs
```diff
@@ -175,38 +188,49 @@ impl<'a> PrimitiveTypeBuilder<'a> {
...
+  /// Sets optional field id and returns itself.
   pub fn with_id(mut self, id: i32) -> Self {
     self.id = Some(id);
     self
   }
 
-  // Creates a new `PrimitiveType` instance from the gathered attributes.
-  // This also checks various illegal conditions and returns `Err` if that that happen.
+  /// Creates a new `PrimitiveType` instance from the gathered attributes.
+  /// Also checks various illegal conditions and returns `Err` if that that happen.
```
sunchao: nit: extra that. happen -> happens.

sadikovi:
I updated the code, can you have a look again? Thanks!

sadikovi:
Thanks!

PR #82, opened 2018-04-10
Integration with parquet-format

sadikovi:
This PR adds `parquet-format` dependency and removes `parquet_thrift`.
Uses parquet-format 2.4.0.

With this patch, users of the crate will no longer need Thrift to build the code.


sadikovi:
This is WIP at the moment, I just want to test if travis passes with my changes, and we need to discuss if this is an appropriate strategy!

sunchao:
Thanks @sadikovi . I agree on the benefits of including `parquet.rs`, but I'm a little hesitant mainly because both parquet-mr and parquet-cpp doesn't do it. This may cause issues if we want to merge into Apache in future.

I'm wondering if there's any workaround on the docs.rs issue. I'll also try to explore on my end before deciding which choice to take.

sunchao:
> Some build step fails on this PR, so coveralls is not reported. I will see if I can do anything about it.
Seems like a transient issue - let me re-launch the build.

Another approach I'm thinking about is to add a feature flag to fetch the `parquet.rs` from the release page. This can be used to skip the Thrift installation, and enable docs.rs. I have a WIP diff - could you take a look?

sadikovi:
I will close this for now, and reopen if we decide to go this route. Current discussion is on #84.

sunchao:
@sadikovi : the `parquet-format` crate is published.

sadikovi:
@sunchao I fixed this PR, had to add `sudo: required` for [kcov problem](https://github.com/travis-ci/travis-ci/issues/9061). Could you review? Thanks!

.travis.yml
```diff
@@ -1,3 +1,5 @@
+dist: trusty
+sudo: required
```
sunchao: hmm. why we don't need this before?

src/lib.rs
```diff
@@ -36,13 +36,13 @@ extern crate brotli;
 extern crate flate2;
 extern crate rand;
 extern crate x86intrin;
+extern crate parquet_format;
```
sunchao: maybe we can also remove some dependencies that were moved to `parquet-format-rs`?

.travis.yml
sadikovi: It looks like a regression in travis. I am not sure why we didn't need it before. I will spend some time figuring out why.

src/lib.rs
sadikovi: Yes, I will remove them. Thanks.

sadikovi: Removed them, now lib.rs should contain only parquet dependencies.

.travis.yml
sadikovi: I checked the travis docs. It looks like whenever you explicitly use `sudo make` or any `sudo <command>`, travis will assume `sudo: required` flag. This is mentioned here: https://docs.travis-ci.com/user/reference/overview/#For-a-particular-.travis.yml-configuration

When sudo is required, then build runs on GCE, otherwise in container on EC2. Here is the comparison:

- master branch (without any changes) [KCOV PASS]: https://travis-ci.org/sunchao/parquet-rs/builds/364441406, "Worker information" mentions GCE.
- build with `sudo: required` [KCOV PASS]: https://travis-ci.org/sunchao/parquet-rs/builds/365997814, "Worker information" also mentions GCE.
- build without any `sudo` keywords [KCOV FAIL]: https://travis-ci.org/sunchao/parquet-rs/builds/365537349, "Worker information" lists EC2.

So we have been running with `sudo: required` all this time, because of this line https://github.com/sadikovi/parquet-rs/blob/master/.travis.yml#L18.

So now we just explicitly state that we require sudo (for kcov issue on travis).

sunchao: I see. Nice finding! Thanks for the detailed explanation.

sadikovi:
@sunchao I rebased. Could you have a look again? Thanks!

sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thank you very much @sunchao! I hope this change will make it easier to use `parquet-rs` crate in different projects!

PR #84, opened 2018-04-11
[WIP] add a flag to download Thrift generated file

sunchao:
An experiment I'm testing. @sadikovi : can you take a look?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,13 @@ flate2 = "0.2"
 rand = "0.4"
 thrift = "0.0.4"
 x86intrin = "0.4.3"
+
+# TODO: only include this when the 'download-thrift-gen' flag is on.
+[build-dependencies]
+reqwest = "0.8.5"
+
+[features]
+download-thrift-gen = []
+
+[package.metadata.docs.rs]
+features = ["download-thrift-gen"]
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,5 @@
+#![feature(custom_attribute)]
+
 use std::env;
 use std::fs;
 use std::process::Command;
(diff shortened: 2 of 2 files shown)


sadikovi:
Changes look good. 

I have an alternative suggestion. I think we could do something similar to parquet-mr here. We can create a separate project/crate and call it `parquet-format`, which will have only one purpose - keeping generated `parquet.rs` and `parquet.thrift`. We could keep version that is the same as parquet-format project.

When we need to update format, we would update `parquet-format-rs` crate with newer version, update its version, and push to crates. Then we would update Cargo.toml for `parquet-rs` to link to a new version, e.g.:
```toml
[dependencies]
parquet-format = "2.3.2"
```

We could even release it as soon as new version of parquet-format comes out and then gradually update parquet-rs as we need to.

Would this approach work?

sunchao:
@sadikovi Yes I like this idea. With this, do you think we should still keep the option of building the `parquet.rs` using Thrift?

sadikovi:
Main motivation for not having Thrift dependency is that `parquet.rs` does not change a lot, only when we update the format.

Since the main constraint is not having parquet.rs in parquet repository, we could just move it to a different repository. I suggest we remove Thrift generation option in this repository - all `parquet.thrift` and `parquet.rs` files will go to `parquet-format-rs` repository.

This is just an idea of working around some of the constraints.

sunchao:
Yes this should be OK. Let's go this route then.

sadikovi:
Thanks! I hope this is not going to break anything. Are you going to create and set up `parquet-format-rs` repository? I suggest we also version it based on actual version of `parquet-format`, so when we link it, we know what version of parquet format we use.

sunchao:
Yes, let me create the repo and set up it. Will add you into it. Meanwhile, I think we can polish #82 to depend on it.

Issue #85, opened 2018-04-13
Add type-safe accessors to Row


andygrove:
When processing parquet files I would like the ability to get values like this:

```rust
let a = row.get_bool().unwrap();
let a = row.get_float().unwrap();
...
```

Currently I have to write pattern matching each time I want to extract a value.

andygrove:
PR: https://github.com/sunchao/parquet-rs/pull/86

andygrove:
I agree. I can make this change but then each accessor will now need to perform two matches - one to check that the Row is a Group and then another match for the contained Row. This is in addition to creating a Vec for each row being read. This is quite expensive.

I think a better approach would be for the row reader to return a trait for reading each row that has accessor methods that can get the value from each column without having to build the intermediate vec. I don't know how big a change that would be though.

andygrove:
I pushed changes to the PR to add this:

```rust
trait RowAccessor {
  fn get_bool(&self, i: usize) -> Result<bool>;
  fn get_byte(&self, i: usize) -> Result<i8>;
  fn get_short(&self, i: usize) -> Result<i16>;
  fn get_int(&self, i: usize) -> Result<i32>;
  fn get_long(&self, i: usize) -> Result<i64>;
  fn get_float(&self, i: usize) -> Result<f32>;
  fn get_double(&self, i: usize) -> Result<f64>;
  fn get_string(&self, i: usize) -> Result<&String>;
  fn get_bytes(&self, i: usize) -> Result<&ByteArray>;
  fn get_timestamp(&self, i: usize) -> Result<u64>;
  fn get_group(&self, i: usize) -> Result<&Vec<(String,Row)>>;
  fn get_list(&self, i: usize) -> Result<&Vec<(Row)>>;
  fn get_map(&self, i: usize) -> Result<&Vec<(Row,Row)>>;
}
```

I then implemented this for Row. Let me know if you like this and I'll finish adding the tests.

andygrove:
I figure with this approach, it will be possible later to have get_row_reader() return an iterator over this perhaps.

sunchao:
> I agree. I can make this change but then each accessor will now need to perform two matches - one to check that the Row is a Group and then another match for the contained Row. This is in addition to creating a Vec for each row being read. This is quite expensive.

[...]

PR #86, opened 2018-04-13
Add type-safe accessors for primitive types in Row

andygrove:


src/record/api.rs
```diff
@@ -55,7 +56,59 @@ pub enum Row {
   Map(Vec<(Row, Row)>) // List of key-value pairs
 }
 
+// Macro to generate type-safe get_xxx methods e.g. get_bool, get_short
+macro_rules! row_primitive_accessor {
+  ($METHOD:ident, $VARIANT:ident, $TY:ty) => {
+    pub fn $METHOD(&self) -> Result<$TY, ParquetError> {
+        match *self {
+            Row::$VARIANT(v) => Ok(v),
+            _ => Err(ParquetError::General(format!("Cannot access {} as {}", self.get_type_name(), stringify!($VARIANT))))
```
sadikovi: I was thinking if it would be easier to just print `self` instead of `self.get_type_name`? One of the pros is that we need not maintain the mapping to string values, but I am happy either way.

sadikovi:
This should fix #85.

src/record/api.rs
andygrove: If we print `self`, we would have a message like `Cannot access 1.2 as Bool` which makes debugging harder?

sadikovi: Fair enough, let's keep type names!

andygrove:
Looks like I need to add tests for the error condition for each type to keep code coverage up. I'll get that done either tonight or tomorrow.

sadikovi:
Well, I think it is okay, but if you want to then it is even better! @sunchao what do you think?

andygrove:
It was actually not much work after all.. I think it's great you have code coverage set up and I'll be learning from you and applying this to DataFusion soon

sunchao:
> Well, I think it is okay, but if you want to then it is even better! @sunchao what do you think?

Yes it would be great if we can cover those too but not a must.

src/record/api.rs
```diff
@@ -55,7 +56,113 @@ pub enum Row {
   Map(Vec<(Row, Row)>) // List of key-value pairs
 }
 
+// Macro to generate type-safe get_xxx methods e.g. get_bool, get_short
+macro_rules! row_primitive_accessor {
+  ($METHOD:ident, $VARIANT:ident, $TY:ty) => {
+    pub fn $METHOD(&self) -> Result<$TY, ParquetError> {
+        match *self {
```
sunchao: nit: we use 2-space indent.

src/record/api.rs
```diff
@@ -55,7 +56,113 @@ pub enum Row {
...
+            Row::$VARIANT(v) => Ok(v),
+            _ => Err(ParquetError::General(format!("Cannot access {} as {}", self.get_type_name(), stringify!($VARIANT))))
```
sunchao: nit: maybe we can use `general_err!` here - a little less verbose
also we limit line width to 90 characters.

src/record/api.rs
```diff
@@ -55,7 +56,113 @@ pub enum Row {
   Map(Vec<(Row, Row)>) // List of key-value pairs
 }
 
+// Macro to generate type-safe get_xxx methods e.g. get_bool, get_short
+macro_rules! row_primitive_accessor {
+  ($METHOD:ident, $VARIANT:ident, $TY:ty) => {
+    pub fn $METHOD(&self) -> Result<$TY, ParquetError> {
```
sunchao: nit: you can import `errors::Result` and replace `Result<$TY, ParquetError> ` with `Result<$TY>`

src/record/api.rs
```diff
@@ -343,4 +450,139 @@ mod tests {
...
+      (Row::Int(3), Row::Float(2.3))
+    ]);
+    assert!(row.get_bool().is_err());
+  }
+
+  #[test]
+  fn test_row_accessors_invalid_message() {
+    assert_eq!(ParquetError::General("Cannot access Float as Bool".to_string()),
+               Row::Float(1.2).get_bool().unwrap_err());
+    assert_eq!(ParquetError::General("Cannot access Float as Str".to_string()), Row::Float(1.2).get_string().unwrap_err());
```
sunchao: nit: line too long.

src/record/api.rs
```diff
@@ -55,7 +56,113 @@ pub enum Row {
...
+  row_primitive_accessor!(get_byte, Byte, i8);
+  row_primitive_accessor!(get_short, Short, i16);
+  row_primitive_accessor!(get_int, Int, i32);
+  row_primitive_accessor!(get_long, Long, i64);
+  row_primitive_accessor!(get_float, Float, f32);
+  row_primitive_accessor!(get_double, Double, f64);
+  row_primitive_accessor!(get_timestamp, Timestamp, u64);
+
+  /// Type-safe accessor for Str type
+  pub fn get_string(&self) -> Result<&String, ParquetError> {
```
sunchao: One thing I'm not quite sure is whether it's better to just panic here, instead of returning `Result`. It seems pretty rare that the error case would happen, given that one always (I think?) will hold the schema of the row before calling these methods.

andygrove: I think panic is pretty extreme. I think the user should decide if they want to panic or not. In non-test code they just need to add a `?` at the end of the line, so it doesn't make much difference to code verbosity.

sunchao:
Hi @andygrove , sorry for the delay! We just refactored the `Row` so that adding accessors to it might become easier. Could you take a look at the latest code and see if the changes can be applied to it (it should not be much difference).

andygrove:
Cool. I am working on it this morning.

andygrove:
@sunchao This is ready for review (as long as you are OK with the 0.009% decrease in code coverage!)

src/record/api.rs
```diff
@@ -41,6 +42,79 @@ pub struct Row {
   fields: Vec<(String, Field)>
 }
 
+impl Row {
+  /// Get then number of fields in this row
```
sadikovi: Should it be `the`?

src/record/api.rs
```diff
@@ -41,6 +42,79 @@ pub struct Row {
...
+        _ => Err(general_err!("Cannot access {} as {}",
+          self.fields[i].1.get_type_name(), stringify!($VARIANT)))
+      }
+    }
+
+  }
+}
+
+impl RowAccessor for Row {
+
```
sadikovi: Could you remove empty line?

src/record/api.rs
```diff
@@ -41,6 +42,79 @@ pub struct Row {
...
+  row_primitive_accessor!(get_float, Float, f32);
+  row_primitive_accessor!(get_double, Double, f64);
+  row_primitive_accessor!(get_timestamp, Timestamp, u64);
+
+  row_complex_accessor!(get_string, Str, String);
+  row_complex_accessor!(get_bytes, Bytes, ByteArray);
+  row_complex_accessor!(get_group, Group, Row);
+  row_complex_accessor!(get_list, ListInternal, List);
+  row_complex_accessor!(get_map, MapInternal, Map);
+
```
sadikovi: Could you remove empty line?

src/record/api.rs
```diff
@@ -71,6 +145,13 @@ pub struct List {
   elements: Vec<Field>
 }
 
+impl List {
+  /// Get then number of fields in this row
```
sadikovi: Should it be `the`?

src/record/api.rs
```diff
@@ -86,6 +167,13 @@ pub struct Map {
   entries: Vec<(Field, Field)>
 }
 
+impl Map {
+  /// Get then number of fields in this row
```
sadikovi: Same here.

andygrove:
@sadikovi I pushed the formatting changes.

Have you thought about adopting cargo fmt? Then you can just have the build fail if the formatting is off and this will no longer be a manual process.

sadikovi:
Thanks for making changes. Looks good, we should merge it!

Yes, we discussed that and at the time, and we needed to apply only a subset of the rules, but  I could not manage to configure it that way. So we decided to review manually and periodically do clean up with rustfmt.

andygrove:
@sunchao Good catch! Fixed.

sunchao:
Thanks @andygrove . Merged.

PR #89, opened 2018-04-15
Refactor Row

sunchao:
This breaks the `Row` struct into two parts: a `RowField`
which represent a single field in a row, and a top-level `Row`
struct which represents the value for a message or struct type.
Only the latter is exposed as public API.

### DIFF
--- a/src/record/api.rs
+++ b/src/record/api.rs
@@ -36,7 +36,34 @@ macro_rules! nyi {
 
 /// Row API to represent a nested Parquet record.
 #[derive(Clone, Debug, PartialEq)]
-pub enum Row {
+pub struct Row {
+  fields: Vec<(String, RowField)>
+}
+
+pub fn make_row(fields: Vec<(String, RowField)>) -> Row {
+  Row { fields: fields }
+}
+
+// TODO: implement `getXXX` for different `RowField`s
+
+impl fmt::Display for Row {
+  fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    write!(f, "{{")?;
+    for (i, &(ref key, ref value)) in self.fields.iter().enumerate() {
+      key.fmt(f)?;
+      write!(f, ": ")?;
+      value.fmt(f)?;
+      if i < self.fields.len() - 1 {
+        write!(f, ", ")?;
+      }
+    }
+    write!(f, "}}")
+  }
+}
+
+/// API to represent a single field in a `Row`.
+#[derive(Clone, Debug, PartialEq)]
+pub enum RowField {
   // Primitive types
   Null,
   Bool(bool),
@@ -50,19 +77,19 @@ pub enum Row {
   Bytes(ByteArray),
   Timestamp(u64), // Timestamp with milliseconds
   // Complex types
-  Group(Vec<(String, Row)>), // Struct, child elements are tuples of field-value pairs
-  List(Vec<Row>), // List of elements
-  Map(Vec<(Row, Row)>) // List of key-value pairs
+  Group(Row), // Struct, child elements are tuples of field-value pairs
+  List(Vec<RowField>), // List of elements
+  Map(Vec<(RowField, RowField)>) // List of key-value pairs
 }
 
-impl Row {
+impl RowField {
   /// Converts Parquet BOOLEAN type with logical type into `bool` value.
   pub fn convert_bool(
     _physical_type: PhysicalType,
     _logical_type: LogicalType,
     value: bool
   ) -> Self {
-    Row::Bool(value)
+    RowField::Bool(value)
   }
 
   /// Converts Parquet INT32 type with logical type into `i32` value.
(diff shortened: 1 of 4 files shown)


sadikovi:
Thanks for making the change! 

First of all, sorry for the long and funky tests in `record` module that you had to change.
And for a conflicting files in this PR.

Looks good, I have a couple questions though.

Now `Row` represents the top level struct and `RowField::Group` represents the some internal struct. But they both would need `getXXX` methods then. Will it result in code duplication? By the way, are `getXXX` methods ones that are part of `RowAccessor` in another PR?

With the current approach we can map `Row` to some other structure, e.g. JSON struct. Is it still possible with this update?

sadikovi:
Another thing - should we rename `Row` to something else? I just think it is a bit weird name for a record:).

sunchao:
Thanks for taking a look @sadikovi .
> Now Row represents the top level struct and RowField::Group represents the some internal struct. But they both would need getXXX methods then. Will it result in code duplication? By the way, are getXXX methods ones that are part of RowAccessor in another PR?

Only `Row` needs to implement the `getXXX` methods, and we don't need to change anything on `RowField::Group` (calling `getStruct` on a `Row` will yield another `Row`). And yes, these `getXXX` methods are the same as in the other PR.

> With the current approach we can map Row to some other structure, e.g. JSON struct. Is it still possible with this update?

The mapping can still happen at the `RowField` level, which not much has changed. We just added a public `Row` API with a few getters.

> Another thing - should we rename Row to something else? I just think it is a bit weird name for a record:).

I'm debating on `Row` or `Record`, and BTW, Spark also [uses](https://spark.apache.org/docs/2.2.1/api/java/org/apache/spark/sql/Row.html) `Row` for similar purpose. `Row` also maps to row group. I don't like very much `RowField` though - maybe we can just call it `Field` or `Column`. I'm not quite sure yet...

src/record/api.rs
```diff
@@ -36,7 +36,34 @@ macro_rules! nyi {
 
 /// Row API to represent a nested Parquet record.
 #[derive(Clone, Debug, PartialEq)]
-pub enum Row {
+pub struct Row {
+  fields: Vec<(String, RowField)>
+}
+
+pub fn make_row(fields: Vec<(String, RowField)>) -> Row {
```
sadikovi: Can you add some doc to the function? Thanks.
Would it be better to keep it as macro, since it is used in different places? Would it be inlined? If so, then it is fine.

src/record/mod.rs
```diff
@@ -17,6 +17,8 @@
 
 //! Contains record-based API for reading Parquet files.
 
-pub mod api;
 pub mod reader;
+mod api;
 mod triplet;
+
+pub use self::api::Row;
```
sadikovi: Should we export `RowField` as well or will it work with just `Row`?

src/record/reader.rs
```diff
@@ -297,7 +298,24 @@ impl Reader {
 
   /// Reads current record as `Row` from the reader tree.
   /// Automatically advances all necessary readers.
+  /// This must be called on the root level reader (i.e., for Message type).
+  /// Otherwise, it will panic.
   fn read(&mut self) -> Row {
```
sadikovi: When do we call `read`? I am just a bit confused:)

src/record/mod.rs
sunchao: I think we should not expose `RowField` (or `Field`). For primitive types, the getters will return Rust native types, such as `i32`, `bool` etc. We also need to define and expose clean-defined interfaces for array and map. The former should also have getters like `getBoolean(rowIndex: usize)` to return the value at a particular row, and the latter should have something like `getKeys()`, `getValues()`, etc. 

I'll take a further look into this. It seems rather verbose if we have to define `getXXX` for every data type on struct, array, etc. I wonder if there's a cleaner way to do it.

src/record/reader.rs
sunchao: The `read` is called on `next()` for both `RowIter` and `ReaderIter`.

src/record/api.rs
```diff
@@ -50,19 +77,19 @@ pub enum Row {
   Bytes(ByteArray),
   Timestamp(u64), // Timestamp with milliseconds
   // Complex types
-  Group(Vec<(String, Row)>), // Struct, child elements are tuples of field-value pairs
-  List(Vec<Row>), // List of elements
-  Map(Vec<(Row, Row)>) // List of key-value pairs
+  Group(Row), // Struct, child elements are tuples of field-value pairs
+  List(Vec<RowField>), // List of elements
```
sadikovi: I guess when it is list of structs, it will `List(Group(Row), Group(Row), ...)`. Is that right?

sunchao: Yes it will be something like `List(Vec(Group(Row), Group(Row),...))`, and the array interface that I proposed above should have something like `getStruct(rowIndex)` which will return a `Row`.

sadikovi: Great, thanks.

Would it be beneficial to also make `List` and `Map` separate, similar to `Row`, so they can define their own set of methods to get length, elements by index, keys and values?

sunchao: I'm thinking about to replace `List<Vec<Row>>` to something like `List<Array>` (we'll need better naming on this) where the `Array` implements the public APIs. Similar for `Map<Vec<Row, Row>>`.  What do you think?

sadikovi: I remember you suggested that some time ago. Yes, sounds like a good idea. You could consider calling them `Struct`, `Array`, `Map`, `Field`, and we would return `Struct` for a record, so it aligns with `get_struct`, `get_array`, and `get_map` methods.

sunchao:
Yes all the TODOs will be fixed later. Will add doc for Map & List, and rebase. Thanks

sadikovi:
@sunchao No worries, thanks for the changes.

PR #94, opened 2018-04-19
Add statistics for column chunk metadata and data page

### LINKED ISSUE #83: Process statistics for column chunk metadata and data page headers
There is a TODO that I think we should complete to have full featured read support. 

https://github.com/sunchao/parquet-rs/blob/master/src/file/metadata.rs#L268
https://github.com/sunchao/parquet-rs/blob/master/src/file/reader.rs#L357

Also take into consideration https://github.com/sunchao/parquet-rs/blob/master/src/file/reader.rs#L332.

Will work on it.

sadikovi:
This PR adds `Statistics` enum in `statistics.rs` that tracks `min`, `max` and `distinct_count`, `nulls`, which are min value for a column in a row group, max value for a column in a row group, optional number of distinct values, and number of nulls respectively. 

`Statistics` is enum that is mapped to all physical types we have, it is considered to be immutable. The actual implementation is in `TypedStatistics<T: DataType>`. This should be used when implementing statistics updates (see the comments below related to the updates).

To support new ordering feature in statistics we added `ColumnOrder` and `SortOrder`. All legacy statistics should be treated as `ColumnOrder::UNDEFINED` and `SortOrder::SIGNED`.

Closes #83


sunchao:
Thanks @sadikovi ! I'll take a look at this soon. Meanwhile, could you take a look at #89 ? I made some changes on that PR.

sadikovi:
Cheers! Yes, I was going to review #89 after your changes - will do asap.

src/file/metadata.rs
```diff
@@ -354,8 +374,391 @@ impl ColumnChunkMetaData {
...
+  UNDEFINED
+}
+
+/// Optional column statistics that are used per row group and per page.
+///
+/// Use `order` field to determine comparison order for a type.
+/// When statistics contain deprecated min/max fields, then comparison is always signed.
+#[derive(Clone, Debug, PartialEq)]
+pub enum Statistics {
+  Boolean { order: ColumnOrder, min: Option<bool>, max: Option<bool>, nulls: u64 },
```
sunchao: wonder if it would be better to have a `Statistics` trait with methods such as `null_count`, `has_nulls`, etc., and a separate struct `TypedStatistics<T>` that implements the former. The latter can implement more methods in future such as `set_min_max`, etc., that are specific to the generic type `T`.

src/file/metadata.rs
```diff
@@ -354,8 +374,391 @@ impl ColumnChunkMetaData {
...
+  Int96 { order: ColumnOrder, min: Option<Int96>, max: Option<Int96>, nulls: u64 },
+  Float { order: ColumnOrder, min: Option<f32>, max: Option<f32>, nulls: u64 },
+  Double { order: ColumnOrder, min: Option<f64>, max: Option<f64>, nulls: u64 },
+  Binary { order: ColumnOrder, min: Option<Vec<u8>>, max: Option<Vec<u8>>, nulls: u64 }
+}
+
+impl Statistics {
+  /// Returns number of nulls for the column.
+  /// Note that even though statistics are for leaf columns, null count also takes into
+  // account null complex types, e.g. lists.
```
sunchao: nit: add a `/`

src/file/metadata.rs
```diff
@@ -354,8 +374,391 @@ impl ColumnChunkMetaData {
...
+      },
+      None => None
+    }
+  }
+
+  /// Returns column order based on physical and logical types.
+  /// This is used when min_value/max_value are set for statistics.
+  ///
+  /// Copied from `parquet.thrift`, see the file for more details on order of values.
+  fn column_order(physical_type: Type, logical_type: LogicalType) -> ColumnOrder {
```
sunchao: perhaps this can be a method for `ColumnDescriptor`?

src/file/metadata.rs
```diff
@@ -354,8 +374,391 @@ impl ColumnChunkMetaData {
...
+      dictionary_page_offset,
+      statistics
     };
     Ok(result)
   }
 }
+
+/// Column order to compare values with statistics min/max.
+#[derive(Clone, Debug, PartialEq)]
+pub enum ColumnOrder {
```
sunchao: nit: change this to `SortOrder` to be consistent with parquet-mr/cpp?

src/file/metadata.rs
```diff
@@ -354,8 +374,391 @@ impl ColumnChunkMetaData {
...
+
+  /// Returns `true` if column contains at least one null, `false` otherwise.
+  pub fn has_nulls(&self) -> bool {
+    self.null_count() > 0
+  }
+
+  /// Method to convert from Thrift definition.
+  /// Note that column type should match statistics, otherwise, there is a risk of
+  /// invalid conversion.
+  pub fn from_thrift(
```
sunchao: do we also need to process `column_orders` from `FileMetaData`?

src/file/metadata.rs
```diff
@@ -354,8 +374,391 @@ impl ColumnChunkMetaData {
...
+  /// Unsigned (depending on physical type either value or byte-wise) comparison.
+  UNSIGNED,
+  /// Comparison is undefined.
+  UNDEFINED
+}
+
+/// Optional column statistics that are used per row group and per page.
+///
+/// Use `order` field to determine comparison order for a type.
+/// When statistics contain deprecated min/max fields, then comparison is always signed.
```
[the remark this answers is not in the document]
sadikovi: Yes, will do.

src/file/metadata.rs
sadikovi: This might have been the wrong name for `ColumnOrder`. This is not a sort order, this is an order which should be used when comparing current value with `min` and `max` from statistics. I might nee to rename it.

src/file/metadata.rs
sadikovi: Yes, I will rename it.

src/file/metadata.rs
sadikovi: Yes, I will update.

src/file/metadata.rs
sadikovi: I will rewrite it, so it will be a method on `TypeDefinedOrder` struct as part of `ColumnOrder` enum. Something like this:
```rust
// Sort order for statistics, to define `min <= value <= max`.
enum SortOrder {
  SIGNED,
  UNSIGNED,
  UNKNOWN
}

// Column order for each leaf column.
enum ColumnOrder {
  TypeDefinedOrder,
  Undefined
}

impl ColumnOrder {
  pub fn sort_order(&self, field: &Type) -> SortOrder;
}
```

I just need to figure out how to handle legacy `min`/`max` values as opposed to `min_value`/`max_value`. I will follow the reference: https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java

src/file/metadata.rs
sadikovi: Based on my comment above, `FileMetaData` will have a method that returns column orders for leaf columns, and sort orders could be extracted from it.

src/file/metadata.rs
sadikovi: Sorry, I am a bit confused. I chose not using `T`, so it is easy to extract relevant statistics a type.

My main concern with typed statistics is that we would need a wrapper similar to column reader. That is why I chose to remove this part and directly implement as enum of different statistics for physical type.

Could you elaborate a little bit more your plan on adding typed statistics? Should we consider statistics as read-only? Thanks.

sadikovi:
@sunchao Would you mind taking a look at my replies? I would like to clarify the changes first. Thanks!

src/file/metadata.rs
sunchao: Yes I was wondering if this could be similar to the `RowGroupStatistics` in the C++ version: https://github.com/apache/parquet-cpp/blob/master/src/parquet/statistics.h#L83, which is readonly, and then TypedStatistics for different types. 

I'm not very familiar with how statistics will be used in the whole codebase though. I think the current approach is fine as well - we can come back and revisit this if needed.

src/file/metadata.rs
sunchao: Sounds good. Maybe also take a look at how C++ handles this: https://github.com/apache/parquet-cpp/blob/master/src/parquet/metadata.cc#L53.

I'll take a look too.

sunchao:
@sadikovi : sure. I've left some comments.

src/file/metadata.rs
sadikovi: Interesting... let me think about this. I do want to expose min/max methods.

src/file/metadata.rs
sadikovi: Thanks! I had a look, and this makes clear when to use type defined order, because I was a bit confused with parquet-mr code. Will update accordingly.

sadikovi:
I will make changes ASAP and let you know. Thanks!

sadikovi:
@sunchao I added typed statistics and refactored `ColumnOrder` and `SortOrder`, and added `column_orders` to `FileMetaData`, to get a sense of what it would look like.

Several issues that I found/introduced:
- Currently if `column_orders` is None, we still create vector with `ColumnOrder::UNDEFINED`. Should we do that, or should we just return None?
- Difficult to extract values with typed statistics, meaning that one would have to downcast `Box<Statistics>` to `TypedStatistics<T>` to get min/max.
- Also difficult to add test cases to check statistics, because of the point above.

Could you advise on the changes/further steps? Thanks.

sunchao:
@sadikovi Thanks for the update. Will take a look soon.

sadikovi:
I am planning to go back to a previous implementation with enum, sorry. I just can’t figure out how to work with traits and go back to typed impls. I mean no offence nor disrespect to anyone, but I feel like enum could be better when reading values from statistics.

Plus, I am updating some other bits, did not do much today, promise to finish it ASAP.

sunchao:
OK no worries. Sorry for the wrong suggestion 😳 . I think the enum approach is fine too.

sadikovi:
@sunchao it is all good, your suggestion was on point. I just got a bit carried away. I will spend some time coming up with a better solution.

sadikovi:
@sunchao I updated the code with a new iteration of `Statistics`. It's change is minor, I updated the PR description, if you would like a bit of an overview.

I also designed Statistics updates, but I have not implemented them here, instead I put the code (not fully complete, but compiles) in gist: https://gist.github.com/sadikovi/a2a5d79f4e4368ea50a5b6b5a0e58cce

The idea is treating `Statistics` as immutable and have `MutableStatisticsBuffer` for updates. But underneath we update `TypedStatistics` with the following traits `StatisticsUpdate` and `AsStatistics` that allow us to update signed/unsigned values and convert back to Statistics.

sadikovi:
I think I need to update `file/mod.rs` and `lib.rs` to include a note about statistics. Will do it ASAP.

src/file/reader.rs
```diff
@@ -170,16 +171,48 @@ impl SerializedFileReader {
...
+        let mut res = Vec::new();
+        for (i, column) in schema_descr.columns().iter().enumerate() {
+          match orders[i] {
+            TColumnOrder::TYPEORDER(_) => {
+              let sort_order = ColumnOrder::get_sort_order(
+                column.logical_type(),
+                column.physical_type()
+              );
+              res.push(ColumnOrder::TYPE_DEFINED_ORDER(sort_order));
+            }
```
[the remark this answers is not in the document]
sadikovi: No, it is not possible at the moment. Since it is compile-safe, we will update it, once there is another option added to column order.

src/file/statistics.rs
```diff
@@ -0,0 +1,498 @@
...
+  }
+
+  /// Returns `true` if min value and max value are set.
+  pub fn has_min_max_set(&self) -> bool {
+    statistics_enum_func![self, has_min_max_set]
+  }
+}
+
+/// Typed implementation for [`Statistics`].
+pub struct TypedStatistics<T: DataType> {
```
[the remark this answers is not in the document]
sadikovi: I added `distinct_count` field to statistics.

src/file/statistics.rs
```diff
@@ -0,0 +1,498 @@
...
+  pub fn min(&self) -> &T::T {
+    self.min.as_ref().unwrap()
+  }
+
+  /// Returns max value of the statistics.
+  pub fn max(&self) -> &T::T {
+    self.max.as_ref().unwrap()
+  }
+
+  /// Whether or not min and max values are set.
```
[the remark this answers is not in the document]
sadikovi: Yes, I update the comments.

src/file/reader.rs
```diff
@@ -170,16 +171,48 @@ impl SerializedFileReader {
...
+    match t_column_orders {
+      Some(orders) => {
+        // Should always be the case
+        assert_eq!(orders.len(), schema_descr.num_columns(),
+          "Column order length mismatch");
+        let mut res = Vec::new();
+        for (i, column) in schema_descr.columns().iter().enumerate() {
+          match orders[i] {
+            TColumnOrder::TYPEORDER(_) => {
+              let sort_order = ColumnOrder::get_sort_order(
```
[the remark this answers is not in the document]
sadikovi: The reason I added method to column order was emphasis that this is a column order feature. When adding to column descriptor, we should consider adding actual `ColumnOrder` to it, because having method `get_sort_order` might be a bit confusing, people would want to get sort order, but maybe it is a format that does not support column order, so the actual sort order might be different from the one returned by the method.

I was thinking about adding `ColumnOrder` field to `ColumnDescriptor`. But this will require some additional changes, that I can make in a separate PR. Let me know if you think it is a good idea.

src/file/statistics.rs
```diff
@@ -0,0 +1,498 @@
...
+
+      Some(res)
+    },
+    None => None
+  }
+}
+
+/// Statistics for a column chunk and data page.
+#[derive(Debug, PartialEq)]
+pub enum Statistics {
```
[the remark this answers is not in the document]
sunchao: No worries. I didn't know the difficulty of extracting min/max values. Let's go with the enum approach.

src/file/reader.rs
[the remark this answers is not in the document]
sunchao: Sounds good.

src/file/reader.rs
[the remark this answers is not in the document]
sunchao: I see. The current approach is fine too. Let's keep it this way then.

sunchao:
Merged. Thanks @sadikovi for the nice work!

sadikovi:
@sunchao Thanks a lot! I do feel like I could have done a better job. I will open other issues if the changes to statistics are necessary. Let me know if there is anything you would like to update.

PR #96, opened 2018-04-22
Add benchmark for compression codecs

### LINKED ISSUE #54: Add benchmark for different compression codecs.

sunchao:
This adds compress and decompress benchmarks for the codes we're using. Results:

test compress_brotli   ... bench:  91,398,584 ns/iter (+/- 18,247,343) = 6 MB/s
test compress_gzip     ... bench:  26,889,660 ns/iter (+/- 12,400,368) = 23 MB/s
test compress_snappy   ... bench:      74,625 ns/iter (+/- 56,679) = 8334 MB/s
test decompress_brotli ... bench:   6,986,044 ns/iter (+/- 3,020,495) = 95 MB/s
test decompress_gzip   ... bench:   6,209,322 ns/iter (+/- 1,494,611) = 106 MB/s
test decompress_snappy ... bench:      38,224 ns/iter (+/- 15,953) = 17367 MB/s

It's strange that Brotli is hopelessly slow but Snappy is unrealistically fast. We may need 
to take a look further.

Closes #54.


benches/codec.rs
```diff
@@ -0,0 +1,110 @@
...
+use std::fs::File;
+
+use parquet::basic::Compression;
+use parquet::file::reader::*;
+use parquet::compression::*;
+
+// 10k rows with types:
+//   binary, int32, int64, int96, boolean, float, double, and fixed_len_byte_array
+// filled with random values.
+const TEST_FILE: &str = "10k.parquet";
```
sadikovi: Could we have a benchmark that is similar to unit tests, where we test with byte arrays?

sunchao: Do you mean generate the byte arrays for different types in the benchmark (instead of reading from the Parquet file)?

sadikovi: If it is possible in this case, yes. Though it looks like benchmark just compresses/decompresses byte arrays, we could make it type specific, similar to encoding/decoding tests. 

Again, only if it is possible and does not affect the benchmark.

sunchao: OK. Let me change it to be type-specific, that is, test column by column. I'd prefer to keep the Parquet file though since it allows us to test compression on different encodings. For instance, it uses PLAIN encoding for float/double, and DELTA_BINARY_PACKING for int32, etc. We can also use it to test data page v1 and v2. The file is generated by a modified version of [DataGenerator](https://github.com/apache/parquet-mr/blob/master/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/DataGenerator.java).

sadikovi: Makes sense to keep the file then. All good.

src/lib.rs
```diff
@@ -132,6 +132,7 @@ extern crate rand;
 extern crate x86intrin;
 extern crate parquet_format;
 extern crate chrono;
+extern crate lazy_static;
```
sadikovi: Do we need to add lazy_static to lib.rs even if it is for benchmarks only?

sunchao: Good point. I think both lazy_static and rand are only used in tests & benchmarks. Let me move them to dev-dependencies.

sunchao:
Merged. Thanks @sadikovi !

PR #97, opened 2018-04-22
Remove features flags that are stablized

sunchao:


sunchao:
Merged. Thanks @sadikovi for the review.

PR #99, opened 2018-04-23
Use column descriptor in convert methods in api.rs

sadikovi:
This PR updates the set of `convert_XYZ` methods in `api.rs` to use `ColumnDescPtr` instead of passing physical and logical types to each method. The reason is providing future access to length, precision and scale, which could be used when converting decimal type.


sadikovi:
Derived from #91.

sadikovi:
@sunchao Could you have a look when you have time? Thanks.
It should be easier to review now.

src/record/api.rs
```diff
@@ -410,102 +392,129 @@ fn convert_timestamp_to_string(value: u64) -> String {
...
 #[cfg(test)]
 mod tests {
+  use std::rc::Rc;
+
   use super::*;
   use chrono;
+  use schema::types::{ColumnDescriptor, ColumnPath, PrimitiveTypeBuilder};
+
+  /// Creates test column descriptor based on provided type parameters.
+  macro_rules! make_column_descr {
```
sunchao: Why this requires a macro but not a normal function?

sadikovi: When adding decimal, I plan to add the second arm that will have signature `$physical_type:expr, $logical_type:expr, $length:expr, $precision:expr, $scale:expr`, so I could use the same name but different parameters. With function I would have to either use length, precision, and scale as 0 for ones that don’t support it, or have different function name. So I chose a macro instead.

sunchao: Ah, I see. Poor man's function overloading. 😀

sunchao:
Thanks @sadikovi ! Merged.

src/record/api.rs
sadikovi: Ahaha:) Yes, you are right, I just thought it would be easier that way.

sadikovi:
@sunchao Thanks for merging!

PR #102, opened 2018-04-24
Remove travis-cargo from CI

sunchao:
It seems we no longer need `travis-cargo` for CI. This replaces it with
`cargo` command and removes its installation.


sunchao:
Merged. Thanks @sadikovi !

PR #104, opened 2018-04-28
Add try_from methods to file reader

sadikovi:
This PR adds a set of `try_from` methods for `SerializedFileReader`, so one can pass path as a string to get reader as a shortcut.

Conversions:
- `TryFrom<File>`
- `TryFrom<&Path>`
- `TryFrom<String>`
- `TryFrom<&str>`

I also updated documentation in `lib.rs`: 
- removed reference to the crate version, so we do not need to update lib.rs for every release. 
- split the massive example into two small ones - makes it more "approachable", IMHO.


src/file/reader.rs
```diff
@@ -203,6 +205,39 @@ impl FileReader for SerializedFileReader {
   }
 }
 
+impl TryFrom<File> for SerializedFileReader {
```
sunchao: can we have a few simple tests for these?

sadikovi: Yes, sure. I will add them.

sadikovi:
Is it okay to use `try_from` method? It is nightly feature, and I have to include attribute 
`#![feature(try_from)]` into every example that uses it. 

As an alternative, I could just add several methods to serialised file reader. Let me know if the approach is okay. Thanks!

sunchao:
I kind of like `try_from` better since it gives you `try_into` which can conveniently convert file/str/etc to a file reader. Also, seems the `try_from` feature is very close to stable (it was stabilized and then reverted).

sadikovi:
@sunchao I added a test, could you have a look when you have time?

sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks @sunchao!

PR #112, opened 2018-05-05
Pass in vec for Codec.compress

sunchao:
This changes `Codec.compress` to accept a vec as a parameter,
instead of allocating & returning a vec every time, therefore allows
reuse of the same allocated vec object.

Also refactored the code to use `e.into()` to convert from io::Error to
ParquetError.


sunchao:
Merged. Thanks @sadikovi .

PR #118, opened 2018-05-09
Replace x86intrin with stable SIMD

### LINKED ISSUE #113: Migrate to stabilized x86/x86_64 SIMD
Currently the SIMD feature requires the `x86intrin` crate and a compile time flag `RUSTFLAGS="-C target-feature=+sse4.2"`. Since [x86/x86_64 SIMD is already stabilized](https://github.com/rust-lang/rust/pull/49664), we should no longer need this crate dependency, and more importantly, enable feature check at run time via the command: `is_x86_feature_detected`.

sunchao:
Currently the SIMD feature requires the x86intrin crate and a compile
time flag `RUSTFLAGS="-C target-feature=+sse4.2"`, which is not
convenient. This replaces the x86intrin dependency with recently
stablized SIMD features, and also replaces the `cfg(target_feature)`
flag with `is_x86_feature_detected`, which dynamically determines
whether SSE 4.2 is supported or not.

Fixes #113.


sunchao:
Merged. Thanks @sadikovi .

PR #121, opened 2018-06-01
Update dependencies versions

sadikovi:
This PR updates versions of dependencies, makes relevant changes, and also makes `rand` changes to be available in tests/benches only, since `rand` crate is marked as dev-dependency.

This also fixes warnings from `quick_error`.


sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks for merging!

PR #122, opened 2018-06-18
Add "To Thrift" conversion to basic types and enums

sadikovi:
This PR adds conversion from `parquet::XYZ` to `XYZ`. This is needed when writing the enums into as part of parquet format structs. This allows to do the following:
```rust
let metadata = parquet::ColumnMetaData {
  type_: column_descr.physical_type().into(),
  ...
}
```

Basically implements `From<XYZ> for parquet::XYZ` for physical type, logical type, page type, compression, encoding, and repetition. I did not add column order, because it needs to be converted differently, since there is no direct one-to-one mapping with format definition.

Added unit tests, and made some small changes, like `tp` into `value` to be consistent in methods for non-type enum names, like `Encoding` or `Compression`.


sunchao:
Interestingly `cargo doc --no-deps` generates lots of warnings, and `cargo coveralls` is failing again (which I think is expected since our travis cargo is not set up properly).

sunchao:
Merged. Thanks @sadikovi !

sadikovi:
@sunchao Thanks!

PR #123, opened 2018-06-22
disallow get_encoder for dict encoding

sadikovi:
This PR updates `get_encoder()` method, similar to its counterpart `get_decoder()` method, to disallow instantiation of `DictEncoder`, because dictionary encoder provides several methods for writing dictionary indices and values, which are not exposed in `Encoder` API.

Another approach would be to just document that dictionary encoder should be instantiated explicitly, if those methods are needed, but then it would defeat having dictionary encoder in the first place. Plus, the current approach is consistent with decoding.

I also updated tests to use `get_encoder` and `get_decoder` methods, similar to decoder tests.

### DIFF
--- a/src/encodings/encoding.rs
+++ b/src/encodings/encoding.rs
@@ -63,7 +63,7 @@ pub fn get_encoder<T: DataType>(
       Box::new(PlainEncoder::new(desc, mem_tracker, vec![]))
     },
     Encoding::RLE_DICTIONARY | Encoding::PLAIN_DICTIONARY => {
-      Box::new(DictEncoder::new(desc, mem_tracker))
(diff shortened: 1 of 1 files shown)


sadikovi:
@sunchao could you review this PR? Adds consistent handling of encoding/decoding. Do you think it would be useful to add any tests for these changes? Thanks!

sunchao:
Test succeeded locally - let me re-trigger the build.

sadikovi:
@sunchao Thanks for the review. I refactored code in both decoding and encoding to have similar test where we initialise encoders/decoders and check the error.

sunchao:
This looks good. Thanks!

PR #124, opened 2018-06-23
Fix travis code coverage

sunchao:
The installation steps for cargo-travis were deleted by mistake.
The CI jobs were succeeding because it was using cache, but recently
started to fail because the cache is gone. This fixes it.


sadikovi:
Well done! I actually did not notice it:) Thanks for fixing the issue.

sunchao:
NP. Seems it is still failing for that `types.h not found` error. Looking...

sadikovi:
I was thinking if it could be related to using `travis-cargo test` command instead of `cargo test`, but it could be something else indeed (looks like that is the only thing that is different between the two conf files).

sunchao:
Yeah the error was strange - seems it is trying to compile against x86 but the docker environment was set up for x86_64.

sunchao:
The failure was caused by https://github.com/SimonKagstrom/kcov/issues/262, and the original commit that caused the issue was reverted. Coveralls worked in the latest build.

sadikovi:
Thanks. Well done!

PR #127, opened 2018-07-01
[WIP] Write support

sadikovi:
This is WIP, to test and discuss the API and implementation. 

_Once design has been approved, I will split work into small chunks and submit them separately._

PR adds low-level write support with the following features:
- Writing pretty much any schema, as long as column writers are provided correct values, definition and repetition levels.
- Support for writing data pages v1 and data pages v2.
- All encodings that are supported for reading are supported in writes.
- All compression levels that are supported for reading are supported in writes.

Limitations:
- Statistics are not supported for writes. This includes sort orders (which is part of statistics API).
- New logical type is not supported, we just write `None`. But for that we will need to update format first.

Testing:
- Not a lot was done for testing, I was just hacking it to make sure that some basic stuff works.
- Tested manually, I added a simple binary to try out the API and write small files, including empty ones. Tested against our `parquet-schema` + `parquet-read` to check metadata and records + Spark.


sunchao:
Thanks @sadikovi ! Very nice work 👍 . I'll try to allocate some time to review this soon.

sadikovi:
It would be good to check the `copy`/`clone` calls that I make in the code.

sadikovi:
I think it might be worth spending some time and building the code to write `Row` into Parquet, so we can test write-read roundtrip.

sunchao:
Hi @sadikovi . Sorry for the delay. I think the WIP PR looks great! Could you split it up into several sub-PRs and submit them for separate review?

> It would be good to check the copy/clone calls that I make in the code.

Will check this in the sub-PRs.

> I think it might be worth spending some time and building the code to write Row into Parquet, so we can test write-read roundtrip.

Hmm not sure if this is necessary, since we already have encoder level round-trip tests. Also, we can do round-trip test for column reader/writer - will that be enough?

sadikovi:
Thanks. I will outline and link issues on what needs to be completed. Yes, column reading/writing should be enough.

PR #131, opened 2018-07-13
Add to_thrift conversion for ColumnChunkMetaData and RowGroupMetaData

sadikovi:
This PR adds `to_thrift` conversion for `ColumnChunkMetaData` and `RowGroupMetaData`, both structs already have `from_thrift` conversion.

I also added builders so metadata can be built outside of `metadata.rs` - this is used in writer when assembling metadata before converting it to Thrift.

Note that `build()` methods return `Result`, `from_thrift()` methods return `Result`, but `to_thrift()` methods return `Self` directly, because we ensure that metadata is built correctly.

Added a few tests to check the conversion.


sadikovi:
@sunchao Could you review this PR as well? Thanks!

src/file/metadata.rs
```diff
@@ -386,4 +474,247 @@ impl ColumnChunkMetaData {
...
+      type_: self.column_type.into(),
+      encodings: self.encodings().into_iter().map(|&v| v.into()).collect(),
+      path_in_schema: Vec::from(self.column_path.as_slice()),
+      codec: self.compression.into(),
+      num_values: self.num_values,
+      total_uncompressed_size: self.total_uncompressed_size,
+      total_compressed_size: self.total_compressed_size,
+      key_value_metadata: None,
+      data_page_offset: self.data_page_offset,
+      index_page_offset: None,
```
sunchao: We convert `index_page_offset` and `statistics` from thrift but not to thrift. Why is that?

src/schema/types.rs
```diff
@@ -497,6 +497,11 @@ impl ColumnPath {
   pub fn string(&self) -> String {
     self.parts.join(".")
   }
+
+  /// Returns slice of parts of the column path.
+  pub fn as_slice(&self) -> &[String] {
```
sunchao: nit: we can implement `AsRef<[String]>` for this purpose too.

src/file/metadata.rs
```diff
@@ -58,6 +58,18 @@ impl ParquetMetaData {
     }
   }
 
+  /// Creates Parquet metadata from file metadata and a list of row group metadata `Rc`s
+  /// for each available row group.
+  pub fn from(
```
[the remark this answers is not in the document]
sadikovi: Sure, will do.

src/file/metadata.rs
sadikovi: I think I forgot to add them. Statistics are not currently supported, that is why I passed `None`. But I will still add the code for it here. Will fix.

src/schema/types.rs
sadikovi: Do you mean adding imp for AsRef trait for ColumnPath and removing `as_slice`?

sadikovi:
@sunchao I updated the code according to your comments. Could you have a look again? Thanks!

src/file/statistics.rs
```diff
@@ -190,6 +190,59 @@ pub fn from_thrift(
...
+    distinct_count: stats.distinct_count().map(|value| value as i64),
+    max_value: None,
+    min_value: None
+  };
+
+  // Get min/max if set.
+  let (min, max) = if stats.has_min_max_set() {
+    let (min, max) = match stats {
+      Statistics::Boolean(ref typed) =>
+        (typed.min().as_bytes().to_vec(), typed.max().as_bytes().to_vec()),
```
sunchao: Can we add methods `min()` and `max()` for `Statistics`? just like in `TypedStatistics`.

sadikovi: Each statistics instance will have its own underlying type. Do you mean `min()` and `max()` returning values as bytes?

sunchao: I see. Yes in that case maybe `min_bytes()` and `max_bytes()`? corresponding methods are also defined in parquet-mr.

sadikovi: Yes, will do. Thanks.

sadikovi:
I updated the code, could you review again? Let me know if there is anything missing!

sunchao:
Merged. Thanks @sadikovi .

sadikovi:
Thanks for merging, @sunchao!

PR #132, opened 2018-07-18
DeltaByteArrayEncoder::flush_buffer() should reset previous byte array

### LINKED ISSUE #47: Fix encodings::encoding::tests::test_byte_array
Seems this test is flaky.

screen shot 2018-02-19 at 9 50 05 pm

sunchao:
If flush_buffer() doesn't reset the previous vector, the first prefix
len the encoder write in the next put() call will potentially be
positive. However, in set_data() DeltaByteArrayDecoder will clear
previous_value, and therefore IndexOutOfBound will occur if we try to
access it using the previous position prefix len.

Fixes #47.


src/encodings/encoding.rs
```diff
@@ -1149,6 +1150,26 @@ mod tests {
     );
   }
 
+  #[test]
+  fn test_delta_byte_array_reuse_encoder() {
```
sadikovi: Would it be possible to name the test so it reflects the problem, e.g. test_delta_byte_array_reset_previous_value or something like that?

sunchao: Sure. I renamed the test to be `test_issue_47` and added a link of the issue to it. Let me know if you think this is OK.

sunchao:
Merged. Thanks @sadikovi for the review :)

PR #135, opened 2018-07-26
Make SerializedFileReader generic to any Reader

gnieto:
Generalized `FileReader` in order to allow multiple inner readers. This
will allow to be able to parse from an in-memory buffer, for example
(via `Cursor<&[u8]>`, for example).

I've also parametrized `FileReader` with `PagerReader` and
`RowGroupReader`, which avoids boxing and some dynamic dispatching.

There are some details pending to be added, but i wanted to know what do you think about this change. My goal was fuzz the crate, and I think it will be helpful that the library can parse from an arbitrary reader (this change is not enough for fuzzing, as there are some `'static` lifetimes that prevents me to fuzz it)

I realize that the signatures are very verbose now, but I think that with a little more of work they can be simplified (for example, seems that `PageReader` or `RowGroupReader`  may be converted into structs, now that are generic to the reader).


src/file/reader.rs
```diff
@@ -39,51 +39,49 @@ use schema::types::{self, SchemaDescriptor, Type as SchemaType};
...
-pub trait RowGroupReader {
+pub trait RowGroupReader<P: PageReader>: Sized {
   /// Get metadata information about this row group.
   fn metadata(&self) -> RowGroupMetaDataPtr;
 
   /// Get the total number of column chunks in this row group.
   fn num_columns(&self) -> usize;
 
-  /// Get page reader for the `i`th column chunk.
-  fn get_column_page_reader(&self, i: usize) -> Result<Box<PageReader>>;
```
gnieto: I saw that this was only called by `get_column_reader`. It's hard to maintain this with the new generic interface. If this is important (as the trait is public), I can try to bring back this method.

src/file/reader.rs
```diff
@@ -197,7 +232,7 @@ impl SerializedFileReader {
   }
 }
 
-impl FileReader for SerializedFileReader {
+impl<R: 'static + ParquetReader> FileReader<SerializedPageReader<R>, SerializedRowGroupReader<R>> for SerializedFileReader<R> {
```
gnieto: The `'static` lifetime is due to `ColumnReaderImpl` being generic over `DataType`

src/file/reader.rs
```diff
@@ -520,7 +566,7 @@ mod tests {
 
   #[test]
   fn test_file_reader_metadata_invalid_start() {
-    let test_file =
+      let test_file =
```
gnieto: Revert

src/record/triplet.rs
```diff
@@ -478,7 +479,7 @@ mod tests {
     expected_rep_levels: &[i16]
   ) {
     let file = get_test_file(file_name);
-    let file_reader = SerializedFileReader::new(file).unwrap();
+    let file_reader: SerializedFileReader<::std::fs::File> = SerializedFileReader::new(file).unwrap();
```
gnieto: Use std::fs::File

src/file/reader.rs
```diff
@@ -39,51 +39,49 @@ use schema::types::{self, SchemaDescriptor, Type as SchemaType};
...
   /// Projected schema can be a subset of or equal to the file schema, when it is None,
   /// full file schema is assumed.
-  fn get_row_iter(&self, projection: Option<SchemaType>) -> Result<RowIter>;
+  fn get_row_iter(&self, projection: Option<SchemaType>) -> Result<RowIter<P, R>>;
 }
 
 /// Parquet row group reader API. With this, user can get metadata information about the
 /// row group, as well as readers for each individual column chunk.
-pub trait RowGroupReader {
+pub trait RowGroupReader<P: PageReader>: Sized {
```
sadikovi: Why do you need `Sized`?

src/file/reader.rs
sadikovi: But it is R that is static, isn't it?

src/file/reader.rs
```diff
@@ -290,13 +315,24 @@ impl RowGroupReader for SerializedRowGroupReader {
       col.compression(),
       col.column_descr().physical_type()
     )?;
-    Ok(Box::new(page_reader))
+    Ok(page_reader)
+  }
+}
+
+
```
sadikovi: Remove new line.

src/file/reader.rs
```diff
@@ -484,6 +520,16 @@ mod tests {
   use parquet_format::TypeDefinedOrder;
   use util::test_common::{get_temp_file, get_test_file, get_test_path};
 
+  #[test]
+  fn test_read_from_memory() {
```
sadikovi: Test is a bit misleading, it should be something `test_read_from_memory_error_not_enough_bytes`, or something like that. I assume it fails because there are not enough bytes for the footer. IMHO, we should explicitly check the error.

src/record/triplet.rs
```diff
@@ -21,6 +21,7 @@ use data_type::*;
 use errors::{ParquetError, Result};
 use record::api::Field;
 use schema::types::ColumnDescPtr;
+use file::reader::RowGroupReader;
```
sadikovi: We normally sort imports in lexicographical order.

gnieto:
> Is there a significant performance impact from boxing and some dynamic dispatching, in particular FileReader and RowGroupReader? Based on what I have seen so far, this is the only way of returning a generic value, unless you want to use enums, but I can be wrong.

It has a cost. It it's significative or not, depend on the amount of times the code is called. Usually,  I use a Box and dynamic dispatching when the return type needs to be resolved at runtime (for example, see `get_decoder`). On the other hand, look at `FileReader::get_row_group`: It always return `SerializedRowGroupReader`, so no boxing should be needed. 

> If going the PR path, we would not need FileReader trait at all, just use struct with generics to specify what reader uses internally.

I will try this. Thanks.

> Have you considered passing input stream instance Box that could be based on a File or Cursor instead? IMHO, this could result in fewer changes overall. For this change, we would have to add more tests to make sure that FileReader<Cursor> works, which is a larger undertaking than making sure that input stream behaviour works for a vector or a file.

I considered, but i prefer to make it generic. It's true that types gets more verbose. But I think that this can be simplified even more (I didn't want to introduce more changes, but I can try to simplify it a little bit more). And I think that code is (a little bit) more idiomatic by using generics instead of boxing.

>  Also SerializedFileWriter will have to rely on File as an input, because of certain semantics; it might result in confusion between FileWriter and FileReader, but this is a minor thing.

I don't find `SerializedFileWriter` on the project. I'm not sure that I understand you.

---

Finally, thanks for taking time to review this. For sure, I will need to add tests, but I wanted to get early feedback and check if you are ok with it before going further.

src/file/reader.rs
sadikovi: My point was whether or not we want to make `ParquetReader` bound to static. I am not sure about it.

gnieto: I think it should not be there. I've tried to get rid of the `'static` lifetime on `DataType`, but it's really hard.

sadikovi:
Okay, it looks like this PR is trying to solve two problems:
- Using generics instead of boxing.
- Having Cursor instead of File.

Would it be possible to address them separately somehow? I would appreciate it, and it would help to clearly see what is added and why.

FYI: We are working on write support, so `SerializedFileWriter` will be added as part of that work.

gnieto:
Yes, I can open two separate PRs. In fact, I squashed this before opening the PR, so it will easy to split this again :)

I'll open a new PR with the generics part once I've some time.

src/file/reader.rs
sunchao: Yes it's hard to remove the `'static` constraint - I think the root cause is `get_decoder` returns boxed instance, which in default is supposed to have `static` lifetime, so that fields associated with `T` within the decoder can live beyond stack space. One possible way is to make the boxed instance have lifetime `'a` but then we'd have to populate `'a` over many places, which is not desirable. 

I think as long as we don't put any borrowed reference in `ParquetReader`, it should be OK.

sunchao:
Thanks @gnieto ! Agree with @sadikovi that it's better to separate this into two PRs. I like the change on `ParquetReader` since in future we'll add more input sources such as HDFS file. Will take a look at the PR for making `FileReader` generic.

PR #137, opened 2018-07-29
Make DataType trait static

sunchao:
There are many places where we require DataType to be static and have
to add the `T: 'static` notation. As DataType is already static in
definition, we should be able to declare it as so and remove all the
notations where it is used.


sunchao:
Merged. Thanks @sadikovi for approving.

Issue #61, opened 2018-03-17
PLAIN encoding for boolean is not following Parquet format


sunchao:
[Parquet encoding format](https://github.com/apache/parquet-format/blob/master/Encodings.md) specifies boolean should be encoded as bit-packed, LSB first. In the current implementation we just encode it as plain 1-bit value.

It is a little confusing though, that the bit-packed encoding, is already deprecated and replaced by RLE/bit-packed hybrid encoding. Nevertheless, parquet-mr follows this specification and still uses the bit-packed encoding, while parquet-cpp uses the same plain encoding as us.

sunchao:
Hmm. What if parquet-mr writes a file with boolean values, and mark the encoding type as `PLAIN`. Seems parquet-rs will init a `PlainDecoder<BoolType>` and try to decode it, which will fail?

sadikovi:
IMHO, parquet-mr will assign correct encoding name to whatever encoding it applies. I really hope it does:)

As long as parquet-mr assigns name of the encoding that matches actual encoding, then we can handle it correctly, regardless how they encode boolean values.

I just wanted to note that we already have the code to handle both and could potentially read both, but as you mentioned, there could be situations when our code might fail to read. Let's keep it open, someone might encounter this issue in the future.

sunchao:
NP. I agree with your comments. Let's keep it open for now.

PR #141, opened 2018-08-05
Level encoder should support writing levels for data page v2

sadikovi:
This PR attempts to update level encoder to support data pages v2 levels, i.e. starting with offset 0 instead of `mem::size_of<i32>()`.

When we write data page v1 levels, we encode the length of the levels in the first 4 bytes of the vector, hence `mem::size_of<i32>()` offset. With data page v2, we explicitly store the length of the levels, so we do not really need to those 4 bytes offset, and we would rather start with 0.

Level decoder already supports reading data page v1 and v2 by having two methods: `set_data`, which reads the first 4 bytes and sets the length, and `set_data_range` that allows to set start and length manually, since we know the start position and data page v2 header has the levels length.

I updated the relevant code and unit tests to reflect that.


sadikovi:
@sunchao could you review this PR? I found this problem when writing tests for column writer. 

Fix works, but IMHO is a bit messy, and I think there is disconnection between level decoder and level encoder when it comes to encoding and decoding for data page v1 and data page v2. Let me know if it is okay, or we should refactor the code somehow. Thanks!

sadikovi:
I think it might be better to have two level encoders and decoders for data page v1 and data page v2, it’s not that much of a code duplication and it makes it clear what functionality is supported for each.

sunchao:
Thanks @sadikovi ! Yeah I'm not sure what's the most elegant way to handle this discrepancy between page v1 and v2. IMO we can just go with this approach and leave potential refactoring as a future work.

For reference, parquet-cpp doesn't seem to handle page v2. On the other hand, parquet-mr doesn't have a level encoder/decoder. Instead it (taking decoding path as example, encoding should be symmetric):

1. if it's v1, initiate value reader based on encoding (RLE or BIT-PACKED), and the RLE (in our case `RleValueDecoder`) will handle the 4-byte length. In our case, we don't have a BIT-PACKED value reader since this is only used for decoding plain boolean values and our plain decoder already does that. BTW: I think #61 is not an issue anymore.

2. if it's v2, initiate a "raw" RLE value decoder, and just pass the definition/repetition levels byte stream to it (unlike our page structure, parquet-mr has separate byte streams for def/rep levels for page v2).

IMO I don't see a clear advantage in either approach.

sadikovi:
Let me try separating v1 from v2 and see if it makes things simpler.

sadikovi:
@sunchao I updated the code to see if it is better to introduce `RLE_V2` for Data Page v2, also did some minor refactoring. Let me know if this approach is better, or if I should revert to the previous approach. Thanks!

src/encodings/levels.rs
```diff
@@ -25,52 +25,70 @@ use errors::{ParquetError, Result};
...
+  }
 }
 
-/// A encoder for definition/repetition levels.
-/// Currently only supports RLE and BIT_PACKED (dev/null) encoding.
-pub struct LevelEncoder {
-  bit_width: u8,
-  encoder: InternalEncoder
+/// Encoder for definition/repetition levels.
+/// Currently only supports RLE and BIT_PACKED (dev/null) encoding, including v2.
```
sunchao: what does `dev/null` here mean?

src/encodings/levels.rs
```diff
@@ -25,52 +25,70 @@ use errors::{ParquetError, Result};
...
+pub enum LevelEncoder {
+  RLE(RleEncoder),
+  BIT_PACKED(u8, BitWriter),
+  RLE_V2(RleEncoder)
```
sunchao: nit nit: can we move `RLE_V2` next to `RLE`?

src/encodings/levels.rs
```diff
@@ -25,52 +25,70 @@ use errors::{ParquetError, Result};
...
 }
 
 impl LevelEncoder {
   /// Creates new level encoder based on encoding, max level and underlying byte buffer.
   /// For bit packed encoding it is assumed that buffer is already allocated with
-  /// 'LevelEncoder::max_buffer_size' method.
+  /// 'levels::max_buffer_size' method.
```
sunchao: nit: can we use `levels::max_buffer_size` here (the backtick syntax).

src/encodings/levels.rs
sadikovi: I do not quite remember, it was something similar to no-op encoding. I will remove it.

sunchao:
Looks good @sadikovi . Will merge shortly after the CI.

sadikovi:
Thanks!

sadikovi:
Thanks for merging!

PR #142, opened 2018-08-08
Generic reader

gnieto:
Generalized FileReader in order to allow multiple inner readers. This
will allow to be able to parse from an in-memory buffer, for example
(via Cursor<&[u8]>, for example).

Note that there are still some issues because of 'static lifetimes (I
guess that it's still due to some type referencing `DataType`), which
limits a little bit how it may be used. For example, this example does
not work at the moment:

```rust
    let v = vec![0, 1, 2, 3];
    let c = Cursor::new(v.as_ref());
    let serialized = SerializedFileReader::new(c).unwrap();

    // compiler refuses to compile due to `v does not life enough`
    let iterator = serialized.get_row_iter(None).unwrap();
```
--- 

I've had to add some generic annotations. If you don't like it, I can Box it (but i think that it adds some usability issues).


src/file/reader.rs
```diff
@@ -18,24 +18,24 @@
...
+use compression::{Codec, create_codec};
 use errors::{ParquetError, Result};
 use file::metadata::*;
 use file::statistics;
 use parquet_format::{ColumnOrder as TColumnOrder, FileMetaData as TFileMetaData};
-use parquet_format::{PageType, PageHeader};
+use parquet_format::{PageHeader, PageType};
 use record::reader::RowIter;
 use schema::types::{self, SchemaDescriptor, Type as SchemaType};
+use std::convert::TryFrom;
```
sadikovi: Do you mind moving std imports back to the top of the file?

src/file/reader.rs
```diff
@@ -92,17 +92,54 @@ pub trait RowGroupReader {
 const FOOTER_SIZE: usize = 8;
 const PARQUET_MAGIC: [u8; 4] = [b'P', b'A', b'R', b'1'];
 
+pub trait Length  {
```
sadikovi: Doc for trait and methods, especially since it is public.

src/file/reader.rs
```diff
@@ -92,17 +92,54 @@ pub trait RowGroupReader {
...
+  fn len(&self) -> u64;
+}
+
+pub trait TryClone: Sized {
```
sadikovi: Doc for trait and methods, especially since it is public.

src/file/reader.rs
```diff
@@ -92,17 +92,54 @@ pub trait RowGroupReader {
...
+
+impl<'a> Length for Cursor<&'a [u8]> {
+  fn len(&self) -> u64 {
+    self.get_ref().len() as u64
+  }
+}
+
+impl<'a> TryClone for Cursor<&'a [u8]> {
+  fn try_clone(&self) -> Result<Self> {
+    Ok(self.clone())
```
sadikovi: I would like to point out that `try_clone` in a File returns a copy of a handler, it does not actually copy a file. In the case of Cursor, Clone would copy the content of the cursor and call `clone` on the inner buffer. So technically, these have different semantics. If this is the expected behaviour then it is okay, otherwise, we might want to revisit it.

src/file/reader.rs
```diff
@@ -197,7 +233,7 @@ impl SerializedFileReader {
   }
 }
 
-impl FileReader for SerializedFileReader {
+impl<R: 'static + ParquetReader> FileReader for SerializedFileReader<R> {
```
sadikovi: Does usage of `'static` fix the example in the PR description?

src/file/reader.rs
```diff
@@ -495,6 +531,20 @@ mod tests {
...
+    let buffer = include_bytes!("../../data/alltypes_plain.parquet");
+    let cursor = Cursor::new(buffer.as_ref());
+
+    let read_from_file = SerializedFileReader::new(File::open("data/alltypes_plain.parquet").unwrap()).unwrap();
+    let read_from_cursor = SerializedFileReader::new(cursor).unwrap();
+
+    let file_iter = read_from_file.get_row_iter(None).unwrap();
+    let cursor_iter = read_from_cursor.get_row_iter(None).unwrap();
+
+    assert!(file_iter.eq(cursor_iter));
```
sadikovi: Does this compare the iterator content?

src/file/reader.rs
gnieto: Note that the inner type is a reference, so cloning the cursor should clone the reference and the position in which the cursor is. If I'm not wrong, this is the same semantic as the file (so, it clones the file handler and keeps the seek position).

src/file/reader.rs
gnieto: No, is what prevents that it works. Not sure if it's due to boxing or DataType. `'static` implies that outlives any other lifetime, which does not happen on the example (as the vector will be dropped when going out of the scope).

src/file/reader.rs
gnieto: Yes: https://play.rust-lang.org/?gist=69620a3ea343915805f0f524900883ce&version=stable&mode=debug&edition=2015

src/file/reader.rs
sadikovi: Although it is true for a slice, you dont forbid any other subtypes. Semantics of TryClone should be stated clearly, when and how it should be implemented.

sadikovi:
With all of these static problems, is it better going bottom up instead, e.g. from changing data types and internals to the parquet reader, and, in some not very distant future, writer. Feel free to open an issue if you think scope is large.

Looks good.

sadikovi:
@gnieto If you have time, could you assess if it would be possible to make file writer generic in a similar fashion? There is no official PR for it, only WIP (yet): #127. Thanks!

gnieto:
@sadikovi Do you mean to do a PR on your branch? If this gets merged, I can try to do it.

sadikovi:
No, I was going to ask to have a look and see if it would be possible. We
can always do it later, once I open PR for file writer... actually let’s do
it then:)

src/file/reader.rs
```diff
@@ -479,9 +524,9 @@ impl<T: Read> PageReader for SerializedPageReader<T> {
 
 #[cfg(test)]
 mod tests {
-  use super::*;
   use basic::SortOrder;
   use parquet_format::TypeDefinedOrder;
+  use super::*;
```
sunchao: nit: can we keep super at the top?

src/file/reader.rs
```diff
@@ -495,6 +540,20 @@ mod tests {
     );
   }
 
+  #[test]
+  fn test_cursor_and_file_has_the_same_behaviour() {
+    let buffer = include_bytes!("../../data/alltypes_plain.parquet");
+    let cursor = Cursor::new(buffer.as_ref());
+
+    let read_from_file = SerializedFileReader::new(File::open("data/alltypes_plain.parquet").unwrap()).unwrap();
```
sunchao: nit: long line.

sunchao:
@gnieto : sorry for the late review. This PR looks good to me.

For the lifetime issue, we can try to look into that but I think it may come with a cost. Loosing the static lifetime constraint may mean that we need to populate the lifetime parameter in many places, and reasoning about lifetime is not modular in Rust, so it's pretty tedious. In addition to that, it can also restrict expressive in another way (e.g., you may no longer store things as fields since the lifetime is bounded by the life of the stack). 

IMO we should evaluate the trade off. For this case in particular, I'm not quite clear how useful it would be to be able to pass borrowed reference into `Cursor`. Also, we can use `Rc` in case that's really needed.

PR #146, opened 2018-08-11
Convert FIXED_LEN_BYTE_ARRAY type with no logical type in record iterator

sadikovi:
This PR adds conversion for FIXED_LEN_BYTE_ARRAY when no logical type is specified. We convert it into `Field::Bytes(byte_array)`. 

Added test to verify conversion.

Issue #140, opened 2018-08-03
Read is much slower than `pyarrow.parquet.read_table`


asdetrefle:
I know it’s still a project on progress but when I am testing it with some 40 files, I found it much slower than the normal way we read it from python. Is that normal?

Each of my file has 21 columns all `double`, containing between 10k-40k rows. In total around 600k rows.

The codes I used to test the speed is just:

```let reader = SerializedFileReader::new(file).unwrap();

let t1 = Instant::now();
let d = reader.get_row_iter(None).unwrap()
     .collect::<Vec<_>>();
let dt = Instant::now() - t1;```

It takes around 3 seconds for all 40 files. But using python it’s around 0.3 seconds. I use:

```t = pq.read_table(file).to_pandas()```

Am I missing something?

sadikovi:
Did you build the library with release profile?

sadikovi:
Plus, I am not sure if the test is comparable, in Rust you are collecting into vector, but in pandas you just return pandas DataFrame.

asdetrefle:
Yes with release. I agree it is not directly comparable but I just wanted to have an idea how much time it costs to load everything into memory and store in a collection. 

But in theory, for example there is a file with 40k row and it took around 300 milliseconds - Nearly 7 microseconds per row. This is slow in absolute terms and is unlikely to be caused by vector allocation.

sadikovi:
We will look into this performance issue, thanks.

asdetrefle:
Thanks. I actually update to 1.30 nightly and retested this morning. It only takes 1.7 seconds now. This is quite strange. I think previously I was using 1.28.

sunchao:
Thanks for reporting this @asdetrefle ! Is there anyway that you can share the files? It will be great if we can reproduce this on own end and start investigating..

asdetrefle:
I'll try to reproduce it on my personal PC and share the data later.

sadikovi:
Yes, the record assembly is very slow. I wrote the benchmark on our test 10k file, here are the results:

file: parquet-rs/benches/record.rs
cargo bench bench_record_reader

```rust
#![feature(test)]
extern crate parquet;
extern crate test;

use parquet::file::reader::{FileReader, SerializedFileReader};
use std::fs::File;
use std::path::Path;
use test::Bencher;

#[bench]
fn bench_record_reader(bench: &mut Bencher) {
  let path = Path::new("/parquet-rs/data/10k-v2.parquet");
[...]

PR #147, opened 2018-08-12
Add benchmark for file reader

sadikovi:
This PR adds benchmark for file reader/record iterator. We test `10k-v2.parquet` file. Current results on my machine are:
```
running 1 test
test record_reader_10k_collect ... bench:  45,931,668 ns/iter (+/- 13,826,371) = 14 MB/s
```

This benchmark is the first step towards understanding the read performance problem, which is tracked in #140.

This PR depends on #146 work, otherwise it fails with not being able to read fixed length byte arrays.


sadikovi:
@sunchao could you please review? Thanks!

PR #149, opened 2018-08-16
Add file writer and row group writer

sadikovi:
This PR is final part of the core implementation of Parquet writes, adds:
- `FileWriter` and its serialised version `SerializedFileWriter`.
- `RowGroupWriter` and its serialised version `SerializedRowGroupWriter`.

I tried to maintain a fairly simple API. The idea and steps are as follows:
1. Create file writer based on file `std::fs::File`, schema `TypePtr` and writer properties `WriterPropertiesPtr` that you want to write.
2. Call `file_writer.next_row_group` to get a mutable reference to a row group writer to add data.
3. Call `row_group_writer.next_column` to add values to a column. Keep iterating until all columns have been written - we do not support partially written row group!
4. After that you can call `close()` on column writer and/or row group writer. They will be closed regardless, before either the next column is requested or the next row group is requested.
5. Keep adding row groups until you have added enough.
6. Call `file_writer.close` to write final metadata to a file.
7. That is it, after that you should have a complete file, so you can check schema and read records from it.

I added a set of unit tests. Please review them as well, they show API in action.


sunchao:
@sadikovi : really sorry that I haven't got chance to look at this yet. Will take a look today.

sadikovi:
No worries, take as much time as needed.

src/file/writer.rs
```diff
@@ -15,16 +15,290 @@
...
+  }
+
+  #[inline]
+  fn has_next_column(&self) -> bool {
+    self.column_index < self.num_columns() && self.row_group_metadata.is_none()
+  }
+
+  #[inline]
+  fn next_column(&mut self) -> Result<&mut ColumnWriter> {
+    if self.row_group_metadata.is_some() {
```
sunchao: nit: we can just call `has_next_column` at the beginning.

src/file/writer.rs
```diff
@@ -15,16 +15,290 @@
...
+
+      self.current_column_writer = Some(column_writer);
+      Ok(self.current_column_writer.as_mut().unwrap())
+    } else {
+      Err(general_err!("No more column writers left"))
+    }
+  }
+
+  #[inline]
+  fn close(&mut self) -> Result<RowGroupMetaDataPtr> {
```
sunchao: what if user calls `close()` before iterating over all columns. Will that cause some error?

src/file/writer.rs
```diff
@@ -15,16 +15,290 @@
...
 use parquet_format as parquet;
+use schema::types::{self, SchemaDescriptor, SchemaDescPtr, TypePtr};
 use thrift::protocol::{TCompactOutputProtocol, TOutputProtocol};
-use util::io::Position;
+use util::io::{FileSink, Position};
+
+/// File writer interface.
+pub trait FileWriter {
+  /// Creates new row group from this file writer.
+  fn create_row_group(&mut self) -> Result<&mut RowGroupWriter>;
```
sunchao: The good thing with this is we cannot call `close` until all the returned `RowGroupWriter` are out of scope. The downside is the restrictiveness: now it's hard to pass the `RowGroupWriter` to some struct which lives beyond the current call stack. Not sure if this will be an issue.

src/file/writer.rs
```diff
@@ -15,16 +15,290 @@
...
+
+  /// Returns true if there are more column writer available, false otherwise.
+  fn has_next_column(&self) -> bool;
+
+  /// Returns the next column writer.
+  /// When no more columns are available, returns `Err`.
+  ///
+  /// This method finalises previous column writer, so make sure that all writes are
+  /// finished before calling this method.
+  fn next_column(&mut self) -> Result<&mut ColumnWriter>;
```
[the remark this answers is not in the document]
sadikovi: I would have to return `Result<Option<&mut ColumnWriter>>`, is it okay?

src/file/writer.rs
```diff
@@ -15,16 +15,290 @@
...
+use file::properties::WriterPropertiesPtr;
 use file::statistics::{to_thrift as statistics_to_thrift};
 use parquet_format as parquet;
+use schema::types::{self, SchemaDescriptor, SchemaDescPtr, TypePtr};
 use thrift::protocol::{TCompactOutputProtocol, TOutputProtocol};
-use util::io::Position;
+use util::io::{FileSink, Position};
+
+/// File writer interface.
+pub trait FileWriter {
```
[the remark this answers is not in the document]
sadikovi: Yes, will do.

src/file/writer.rs
```diff
@@ -15,16 +15,290 @@
...
+  total_num_rows: u64,
+  row_groups: Vec<RowGroupMetaDataPtr>,
+  row_group_writer: Option<SerializedRowGroupWriter>,
+  is_closed: bool,
+}
+
+impl SerializedFileWriter {
+  /// Creates new file writer.
+  pub fn new(mut file: File, schema: TypePtr, properties: WriterPropertiesPtr) -> Self {
+    Self::start_file(&mut file).expect("Start Parquet file");
```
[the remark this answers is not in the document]
sadikovi: I will update the `new` method to return `Result<Self>` similar to `FileReader` and improve the error message.

src/file/writer.rs
sadikovi: I am going to make it iterator-like, this condition will be updated.

src/file/writer.rs
sadikovi: Yes, it will raise an error, because there will be mismatch between expected columns (must be greater than 0) in the schema and actual number of columns written.

src/file/writer.rs
sadikovi: Yes, I had the same thought, but it looks like this is one of the few options to keep track of row group writers ourselves. Do you have other ideas? Should we just keep a reference count of row group writers and check later when asking for the next one?

src/file/writer.rs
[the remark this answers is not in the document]
sadikovi: I updated the method docs to emphasise what you mentioned. There is a follow-up sub-task to update all of the documentation related to writes after we have merged the implementation.

sadikovi:
@sunchao I addressed your comments, but might have missed some. Would you mind doing another review pass, including tests (I feel they are a bit overcomplicated)? And could you also give your thoughts on the questions I raised in my first comment?

Thanks!

src/file/writer.rs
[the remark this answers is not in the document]
sunchao: OK. Sounds good then.

src/file/writer.rs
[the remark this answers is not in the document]
sunchao: Yes, I think this will be better. We should differentiate the error case and the case where there's no next.

src/file/writer.rs
sunchao: Yes, I think another approach is to return a unique reference to the `RowGroupWriter`:
```
fn create_row_group(&mut self) -> Result<RowGroupWriter>;
```

And a separate function to explicitly close it:
```
fn close_row_group(&mut self, RowGroupWriter) -> Result<()>;
```

This will call `finalise_row_group_writer` to close the row group writer and set some flag in the file writer indicating that the current row group writer has been closed.  We can do similar things for `RowGroupWriter` to make sure all column writers are handed back before the `close` is called. 

Let me know what you think. This is just some rough thought.

sadikovi: Sounds good, I will make changes, see what it looks like, and let you know. Thanks!

src/file/writer.rs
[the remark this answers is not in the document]
sadikovi: Yes, I changed the code to return `Result<Option<Something>>` so looks like iterator and it is similar to how we read pages.

sadikovi:
@sunchao I made the changes to the API as per your suggestion. Can you have a look? Thanks!

Now we have `next_XYZ` and `close_XYZ`, we will raise an error if next method is called before close method is called for the previous XYZ.

sunchao:
@sadikovi : the latest PR looks good to me in general. However, the test `test_column_writer_check_metadata` is failing. Can you fix that?

sadikovi:
Yes, thanks. I just checked the tests, will fix shortly, make some other
minor changes and ping you again.

sadikovi:
@sunchao I fixed the test - it failed because I added test on encodings in this branch, but in master we have updated to add levels encoding as well, so it was failing with missing encoding for levels.

Should be all good now, would appreciate if you could look over the code again. There is also an open question about limiting row groups somehow, either based on size or number of rows - currently we do not limit them in any way. We could also add this as a follow-up issue.

sunchao:
> Should be all good now, would appreciate if you could look over the code again. There is also an open question about limiting row groups somehow, either based on size or number of rows - currently we do not limit them in any way. We could also add this as a follow-up issue.

Yes we should have some config properties for this. Good for a follow-up. :)

sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks!

PR #150, opened 2018-08-17
Fix written rows metric in column writer

sadikovi:
This PR addresses a problem in column writer when writing nullable non-repeated values. In this case the number of rows is the number of definition levels (at most number of values), but we were passing the actual number of values to write.

I also removed `if` statements in the loops in the hopes of improving performance.

Added test that verifies the correctness of the fix.


sadikovi:
@sunchao Can you review this PR? Thanks!

src/column/writer.rs
```diff
@@ -298,9 +298,7 @@ impl<T: DataType> ColumnWriterImpl<T> {
       let levels = def_levels.unwrap();
       num_values = levels.len();
       for &level in levels {
-        if level == self.descr.max_def_level() {
-          values_to_write += 1;
-        }
+        values_to_write += (level == self.descr.max_def_level()) as usize & 1;
```
sadikovi: @sunchao I just realised, should we remove `& 1`, since it is not really needed - boolean is either 0 or 1. I can open PR to fix it.

sunchao: Yes we should not need that `&`. Feel free to add a PR to fix it. :)

PR #151, opened 2018-08-18
Handle repeated fields with no annotation in record reader

sadikovi:
This PR fixes the issue of not being able to read files that contain specifically defined arrays/lists that not marked as `LIST` logical type. For example,
```
message user {
  REQUIRED INT32 id;
  OPTIONAL group phoneNumbers {
    REPEATED group phone {
      REQUIRED INT64 number;
      OPTIONAL BYTE_ARRAY kind (UTF8);
    }
  }
}
``` 

The current master branch code panics when reading this file, because we convert all of the group types as structs, but we should convert repeated `phone` field into a list of `phone` elements (number, kind). 

Spark returns the following result when reading the file:

```
+---+-------------------------------------------------------------------------+
|id |phoneNumbers                                                             |
+---+-------------------------------------------------------------------------+
|1  |null                                                                     |
|2  |null                                                                     |
|3  |[WrappedArray()]                                                         |
|4  |[WrappedArray([5555555555,null])]                                        |
|5  |[WrappedArray([1111111111,home])]                                        |
|6  |[WrappedArray([1111111111,home], [2222222222,null], [3333333333,mobile])]|
+---+-------------------------------------------------------------------------+
```

It seems that we forgot to handle a special case that is mentioned here https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala#L73.

I added the code to handle this case and also added new file with such schema and test case for it.


sadikovi:
@sunchao Could you review this PR? Thanks!

PR #152, opened 2018-08-19
Add RLE encoding to the list of encodings when writing metadata

sadikovi:
This PR adds level encoding to the list of Thrift encodings, when we write column chunk metadata. I forgot to include it in my previous PR. It looks like there is no enforced order on encodings - every system out there tries writing in whatever order they want, and format does not enforce it. Here we are following parquet-cpp.

Also removed redundant `&` operations that I forgot to remove in my previous patch.


sadikovi:
@sunchao Can you review? Thanks.

PR #153, opened 2018-08-19
Format long Float and Double fields

sadikovi:
This PR addresses the issue in record reader of displaying long values of Field::Float and Field::Double. When value is long (`value > 1e19 || value < 1e-15`), we display it in scientific notation, otherwise the default one.

I found this definition of "long" values here: https://github.com/maciejhirsz/json-rust/pull/32/files#diff-d0b3721173a6539188b808c57c0b4b08R54


sadikovi:
@sunchao Could you review this PR? Thanks!

PR #154, opened 2018-08-20
Add a new data file

asdetrefle:
Added a new data file for benchmark. It contains 41k rows and 21 columns. It is the stocks bid/ask price and qty generated from simulation.


sadikovi:
How large is the file? If it is too large we could generate it before each benchmark, as we have (almost) all of the functionality to do so now.

Should we update the benchmark as well and post results on PR? 

Is it possible to get a breakdown of time spent like the one you mentioned in the related issue?

sadikovi:
Should be alright, IMHO.

PR #155, opened 2018-08-23
add two benchmark test with new data file

asdetrefle:
The second benchmark test is a for low level column reader call on the new data file. This is theoretically the lower bound of read time needed. We also should aim for this when we implement `Column` and `Table`. Now we can see huge potential for the current implementation:

```
test record_reader_10k_collect   ... bench:  17,203,101 ns/iter (+/- 109,298) = 38 MB/s
test record_reader_stock_collect ... bench: 116,401,996 ns/iter (+/- 1,467,345) = 11 MB/s
test record_reader_stock_column  ... bench:   6,892,490 ns/iter (+/- 26,985) = 187 MB/s
```


benches/reader.rs
```diff
@@ -23,8 +23,13 @@ use std::fs::File;
 use std::path::Path;
 
 use parquet::file::reader::{FileReader, SerializedFileReader};
+use parquet::schema::types::ColumnPath;
+use parquet::column::reader::{ColumnReader, get_typed_column_reader};
```
sadikovi: We normally sort imports, e.g. 
```rust
use parquet::column::*;
use parquet::data_type::*;
use parquet::file::*;
use parquet::schema::*;
```

And `std::*` go before other imports.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
...
+
+  bench.bytes = len;
+  bench.iter(|| {
+    let iter = parquet_reader.get_row_iter(None).unwrap();
+    let _ = iter.collect::<Vec<_>>();
+  })
+}
+
+#[bench]
+fn record_reader_stock_column(bench: &mut Bencher) {
```
sadikovi: I was thinking about calling it `record_reader_stock_simulated_column`.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
     let _ = iter.collect::<Vec<_>>();
   })
 }
+
+#[bench]
+fn record_reader_stock_collect(bench: &mut Bencher) {
```
sadikovi: I was thinking about calling it `record_reader_stock_simulated_collect`.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
...
+  /// WARNING THIS BENCH IS INTENDED FOR THIS DATA FILE ONLY
+  /// COPY OR CHANGE THE DATA FILE MAY NOT WORK AS YOU WISH
+  let path = Path::new("data/stock_simulated.parquet");
+  let file = File::open(&path).unwrap();
+  let len = file.metadata().unwrap().len();
+  let parquet_reader = SerializedFileReader::new(file).unwrap();
+
+  let descr = parquet_reader.metadata().file_metadata().schema_descr_ptr();
+  let num_row_groups = parquet_reader.num_row_groups();
+  let batch_size = 4096;
```
sadikovi: Record reader currently uses batch size 256. It is not like it is going to make a difference, just FYI.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
...
+      let mut def_levels = Some(vec![0; batch_size]);
+      let mut rep_levels = None::<Vec<i16>>;
+
+      for (_t, col_reader) in readers.into_iter() {
+        match col_reader {
+          r @ ColumnReader::Int64ColumnReader(..) => {
+            let mut data_collected = Vec::with_capacity(num_rows);
+            let mut val = vec![0; batch_size];
+            let mut rr = get_typed_column_reader::<Int64Type>(r);
+            while let Ok((rs, _)) = rr.read_batch(batch_size,
```
sadikovi: Can you use `values_read` and `levels_read` instead? Otherwise, it is not clear what `rs` is. Same applies to `rr` - just call it `typed_reader` or `typed_column_reader`.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
...
+                                                  def_levels.as_mut().map(|x| &mut x[..]),
+                                                  rep_levels.as_mut().map(|x| &mut x[..]),
+                                                  &mut val)
+            {
+                data_collected.extend_from_slice(&val);
+                if rs < batch_size { break }
```
sadikovi: Hmmm... I am not sure. If values are required then this is okay (assuming that internally we always return the max number of values == requested), though you would potentially read less than record reader. If there are null values, then we should check levels_read as well.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
...
+        let col_path = ColumnPath::new(vec![field.name().to_owned()]);
+        let orig_index = *paths.get(&col_path).unwrap();
+        let col_reader = row_group_reader.get_column_reader(orig_index).unwrap();
+        readers.push((field.name().to_owned(), col_reader));
+      }
+
+      let mut def_levels = Some(vec![0; batch_size]);
+      let mut rep_levels = None::<Vec<i16>>;
+
+      for (_t, col_reader) in readers.into_iter() {
```
sadikovi: It looks like you do not use `t` - might as well remove it from `readers` vector altogether.

benches/reader.rs
asdetrefle: I did check that `data_collected` has a length of `41000` which means the reader read till the end. As I mentioned in the beginning, we should not expect the same code works for another data file immediately because there are a lot of other checks I skipped.

benches/reader.rs
asdetrefle: with 256 it's 20% slower.

sadikovi: Then change it so it is comparable!

benches/reader.rs
sadikovi: Well, it is just not clear from the code. Add explicit assertion and/or comment about what is going on in the benchmark.

benches/reader.rs
```diff
@@ -38,3 +43,95 @@ fn record_reader_10k_collect(bench: &mut Bencher) {
...
+      let mut def_levels = Some(vec![0; batch_size]);
+      let mut rep_levels = None::<Vec<i16>>;
+
+      for col_reader in readers.into_iter() {
+        match col_reader {
+          r @ ColumnReader::Int64ColumnReader(..) => {
+            let mut data_collected = Vec::with_capacity(num_rows);
+            let mut val = vec![0; batch_size];
+            let mut typed_reader = get_typed_column_reader::<Int64Type>(r);
+            while let Ok((values_read, _levels_read)) = typed_reader.read_batch(batch_size,
```
sadikovi: nit: IMHO, it is easier to read when it is formatted like this:

```rust
while let Ok((values_read, _levels_read)) = typed_reader.read_batch(
  batch_size,
  def_levels.as_mut().map(|x| &mut x[..]),
  rep_levels.as_mut().map(|x| &mut x[..]),
  &mut val
) 
{
  ...
}
```

sadikovi:
Thanks for the benchmarks! Now we need to figure out what we are going to do to improve the performance.

sadikovi:
LGTM

sunchao:
LGTM too. Thanks @asdetrefle !

@sadikovi : I think you should be able to merge this. Can you try it?

sadikovi:
@sunchao Yes, I can merge as well! Thanks for the review and sign off. Now, I can help with merging PRs too. @asdetrefle - great job!

PR #158, opened 2018-09-07
Add statistics to parquet-schema command display

### LINKED ISSUE #156: Parquet-schema command should also report statistics if available
When debugging Parquet files I found that it can be useful to see the column statistics (if available), e.g. when trying to figure out if row group filtering is happening in Spark.

Would be great if we could print statistics (min, max, nulls, maybe column order) in addition to all of the metadata the command already displays.

sadikovi:
This PR adds `statistics` value to display for a column when running `parquet-schema` command with additional information display (`parquet-schema /path/to/file true`). Example of some prints are below:

```
column 0:
--------------------------------------------------------------------------------
column type: INT32
column path: "b_struct.b_c_int"
...
statistics: {min: N/A, max: N/A, distinct_count: N/A, null_count: 8, 
min_max_deprecated: true}

column 0:
--------------------------------------------------------------------------------
column type: BYTE_ARRAY
column path: "first"
...
statistics: {min: [97], max: [97], distinct_count: N/A, null_count: 0, 
min_max_deprecated: true}

column 10:
--------------------------------------------------------------------------------
column type: INT96
column path: "timestamp_col"
...
statistics: N/A
```

Unfortunately, statistics string can be a bit long, for example:
```
column 1:
--------------------------------------------------------------------------------
column type: DOUBLE
column path: "bp2"
...
dictionary page offset: 10788
statistics: {min: 69.05000000000001, max: 70.60000000000001, distinct_count: N/A, 
null_count: 0, min_max_deprecated: false}
```

Closes #156


sadikovi:
@sunchao could you review this PR? Let me know if I should make any changes or add more tests. Thanks!

sunchao:
Looks good. I'm wondering if it's better to make this optional, i.e., only print it when an extra argument such as `-statistics` is specified. Thought?

sadikovi:
That is a good idea, though we would have to provide `true` and `-statistics` flags and they should work together. My understanding was that if user provided `true` we show everything.

Slightly off-topic: I was thinking if we should improve our CLI tools, for example, have better handling of parameters and help display like `cargo --help` and have different modes for parquet-schema - `--short` for just schema, `--ext` for bunch of other properties and `--all` for everything.

I can add just an extra option for statistics or I can extend to support different options, like I mentioned above. Let me know what you think is better - I am happy to change either way!

sunchao:
Agree that the CLI tool needs improvement - I was also thinking that we can enable colored output as well as making the output more formatted, etc. 

I think we can just re-use the `true` option for everything now, and leave the aforementioned improvements as a followup. :)

PR #161, opened 2018-09-12
Add reader properties

sadikovi:
This PR adds ReaderProperties struct, similar to WriterProperties used for writing. Currently this struct only contains batch_size, which is used to be hard-coded.

Following is done:

- Added ReaderProperties into src/file/properties.rs
- Updated SerializedFileReader::new() method to take additional ReaderProperties
- Updated SerializedRowGroupReader::new() method to take additional ReaderProperties
- Updated FileReader and RowGroupReader traits to have properties() method that returns reference-counted reader properties
- Updated row iterator to take tree builder and pass batch size from properties

This is a fairly big change and has implications of incompatibility of API with prior versions.
- Pros: having configurable batch size and potentially other reader properties.
- Cons: API change, people do not change it themselves, it might be enough to just increase default batch size to 1024/4096.


sadikovi:
@sunchao Could you review this PR and let me know if we should keep the change or postpone it for now? This is one of the (optional) items on write-support issue.

sunchao:
Thanks @sadikovi ! Just curious: do you have in mind what other reader properties we will add in future? I checked parquet-cpp/parquet-mr and didn't find many properties that people need to tune on the reader side.

Issue #160, opened 2018-09-11
Failed to install parquet using "cargo install"


sadikovi:
Rust version: rustc 1.30.0-nightly (551244f05 2018-09-10)
Parquet version: 0.3 (the latest is released on crates.io)
It was working okay in 1.28. I am not sure why it needs to compile files that are used for testing only.

When I run `cargo install parquet`, I get the following error:
```
$ cargo install --force parquet
    Updating registry `https://github.com/rust-lang/crates.io-index`
  Installing parquet v0.3.0                                                                                                                                                   
 Downloading lz4-sys v1.8.2
   Compiling serde v1.0.78                                                                                                                                                    
   Compiling byteorder v1.2.6
   Compiling unicode-width v0.1.5
   Compiling libc v0.2.43
   Compiling num-traits v0.2.5
   Compiling gcc v0.3.54
   Compiling glob v0.2.11
   Compiling remove_dir_all v0.5.1
   Compiling bitflags v0.5.0
   Compiling version_check v0.1.4
   Compiling cfg-if v0.1.5
   Compiling cc v1.0.24
   Compiling void v1.0.2
   Compiling num-integer v0.1.39
   Compiling byteorder v0.5.3
   Compiling integer-encoding v1.0.5
   Compiling alloc-no-stdlib v1.3.0
   Compiling try_from v0.2.2
   Compiling rustc-serialize v0.3.24
   Compiling x86intrin v0.4.5
   Compiling quick-error v1.2.2
   Compiling getopts v0.2.18
   Compiling log v0.4.5
   Compiling base64 v0.5.2
   Compiling rand v0.4.3
   Compiling num_cpus v1.8.0
   Compiling time v0.1.40
   Compiling lazy_static v1.1.0
   Compiling unreachable v1.0.0
   Compiling brotli-decompressor v1.3.1
   Compiling log v0.3.9
   Compiling pulldown-cmark v0.0.8
   Compiling threadpool v1.7.1
   Compiling tempdir v0.3.7
   Compiling lz4-sys v1.8.2
   Compiling miniz-sys v0.1.10
   Compiling num-traits v0.1.43
   Compiling thrift v0.0.4
   Compiling snap v0.2.5
   Compiling ordered-float v0.5.1
   Compiling brotli v2.5.0
   Compiling chrono v0.4.6
   Compiling num-bigint v0.1.44
   Compiling skeptic v0.9.0
   Compiling flate2 v0.2.20
[...]

PR #162, opened 2018-09-14
Update cargo dependencies

sadikovi:
This small PR updates dependencies to the latest versions.


sadikovi:
@sunchao Do you think that we should use `~` to restrict versions in Cargo.toml (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)?

sunchao:
Thanks @sadikovi . The PR looks good to me. Regarding the "tilde" versions, I think we may not need it for now as #160 may just be a Rust nightly upgrade issue. Also, I checked some other major Rust projects and they all use the caret versions.

PR #163, opened 2018-09-14
Update batch size in record reader 

sadikovi:
This PR updates batch size from 256 to 1024.
I also removed TODO for reader properties.


sadikovi:
@sunchao could you review? Thanks!

PR #171, opened 2018-10-11
Fix build error when fetching git hash

### LINKED ISSUE #170: Depending on parquet 0.4.0 from crates errors with git command failure
Looks like the published 0.4.0 version of parquet is erroring during build, doesn't matter if it's release or debug.

The build script was added in https://github.com/sunchao/parquet-rs/pull/128

Here's the failing Cargo.toml dependency and then the error which happens during build:

```
[dependencies]
parquet = "0.4.0"
```

error:

```
error: failed to run custom build command for `parquet v0.4.0`
process didn't exit successfully: `/Users/xlange/code/rust/dracula/target/debug/build/parquet-05fa2decd80b738e/build-script-build` (exit code: 101)
--- stdout
Running: `"git" "rev-parse" "HEAD"`

--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "Failed: `\"git\" \"rev-parse\" \"HEAD\"` (exit code: 128)"', libcore/result.rs:1009:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get
             at libstd/panicking.rs:476
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:390
   6: std::panicking::try::do_call
             at libstd/panicking.rs:325
   7: core::ptr::drop_in_place
             at libcore/panicking.rs:77
   8: core::result::unwrap_failed
             at libcore/macros.rs:26
   9: <core::result::Result<T, E>>::unwrap
             at libcore/result.rs:808
  10: build_script_build::main
             at ./build.rs:23
  11: std::rt::lang_start::{{closure}}
             at libstd/rt.rs:74
  12: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:310
  13: macho_symbol_search
             at libpanic_unwind/lib.rs:102
  14: alloc::alloc::box_free
             at libstd/panicking.rs:289
             at libstd/panic.rs:392
             at libstd/rt.rs:58
  15: std::rt::lang_start
             at libstd/rt.rs:74
  16: build_script_build::run
```

this is fixed by changing parquet to a git source, where the `git` command will return a value.

```
[dependencies]
parquet = { git = "https://github.com/sunchao/parquet-rs.git", branch = "master" }
```

looks like the build.rs (https://github.com/sunchao/parquet-rs/blob/master/build.rs#L20-L28) needs to be changed so it doesn't unwrap and just fails gracefully?

sunchao:
When the parquet-rs crate is used in a third-party library, the "git"
command in the build.rs will fail. This fixes it by skipping the git
hash and just use "parquet-rs version <version-no>" as the CREATED_BY
string. This should be OK since a crate version should always map to
a unique git hash.

Fixes #170

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -20,10 +20,12 @@ use std::process::Command;
 fn main() {
   // Set Parquet version, build hash and "created by" string.
   let version = env!("CARGO_PKG_VERSION");
-  let git_hash = run(Command::new("git").arg("rev-parse").arg("HEAD")).unwrap();
-  let created_by = format!("parquet-rs version {} (build {})", version, git_hash);
+  let mut created_by = format!("parquet-rs version {}", version);
+  if let Ok(git_hash) = run(Command::new("git").arg("rev-parse").arg("HEAD")) {
+    created_by.push_str(format!(" (build {})", git_hash).as_str());
+    println!("cargo:rustc-env=PARQUET_BUILD={}", git_hash);
+  }
   println!("cargo:rustc-env=PARQUET_VERSION={}", version);
-  println!("cargo:rustc-env=PARQUET_BUILD={}", git_hash);
   println!("cargo:rustc-env=PARQUET_CREATED_BY={}", created_by);
 }
 


sadikovi:
One of the ways of fixing it is having a script that updates the hardcodes the version during the release process.

sunchao:
Sure @sadikovi . Thanks.

> One of the ways of fixing it is having a script that updates the hardcodes the version during the release process.

What's the issue with the current approach? we are getting the version from `CARGO_PKG_VERSION`. Is that not the right approach?

sadikovi:
Hello. You are correct - it is the right approach, but parquet version is generally “parquet x.y.z (build abc123)”. Having just the version is correct, though it is not full version, and there could be code written in other processing engine that relies on it. Your fix is fast and on point, thank you for fixing it. But long term we should add something that injects full version of parquet - I will try coming up with the least painful for development and release approach.

sunchao:
I see. You mean we should still keep the `(build abc123)` part for compatibility reason, right? That makes sense. Thanks!

sadikovi:
It is not necessary, but this is what cpp and mr do. We will fix it later.

sunchao:
@sadikovi : I'm going to release 0.4.1 with this fix. After you have a more complete fix we can do another minor release. Is that OK?

sadikovi:
All good, thanks.

Issue #178, opened 2018-10-30
Panic when reading parquet files


gnieto:
I downloaded some sample files from https://github.com/gitential/datasets/tree/master/antirez-redis and I'm not able to load the schema or read the file.

**Example file**: https://s3.amazonaws.com/gitential-datasets/antirez-redis/tags.parquet
**Branch**: master
**Command**:  RUST_BACKTRACE=1 cargo run --bin parquet-schema -- tags.parquet
**Output**:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `6`', src/file/metadata.rs:234:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:476
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:390
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:345
   7: parquet::file::metadata::RowGroupMetaData::from_thrift
             at src/file/metadata.rs:234
   8: <parquet::file::reader::SerializedFileReader<R>>::parse_metadata
             at src/file/reader.rs:197
   9: <parquet::file::reader::SerializedFileReader<R>>::new
             at src/file/reader.rs:150
  10: parquet_schema::main
             at src/bin/parquet-schema.rs:85
  11: std::rt::lang_start::{{closure}}
             at libstd/rt.rs:74
  12: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:310
  13: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102
  14: std::rt::lang_start_internal
             at libstd/panicking.rs:289
             at libstd/panic.rs:392
             at libstd/rt.rs:58
  15: std::rt::lang_start
             at libstd/rt.rs:74
  16: main
  17: __libc_start_main
  18: _start
```

sadikovi:
Interesting. Let me check the file.

sadikovi:
[...]

PR #181, opened 2018-11-02
Support unsigned types and timestamp millis in record reader

sadikovi:
This PR adds support for the following logical types for the record reader API:
- INT32 physical type
  - UINT_8
  - UINT_16
  - UINT_32
- INT64
  - TIMESTAMP_MILLIS
  - UINT_64

I named the new fields as UByte, UShort, UInt and ULong. I know, these names are not the best, but I can't change the existing API to make them all conform to either INT_XY or UINT_XY.

I also updated accessors' methods and updated unit tests.

Relates to #178.


sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks! It is not the best change, but I could not find better names for the new types.

PR #182, opened 2018-11-02
Patch schema to handle root node repetition and num_children as 0

sadikovi:
This PR adds workarounds for some of the parquet-cpp files that have repetition level set for schema root node and primitive fields having 0 as number of children.

Neither of the aforementioned behaviours is part of the format definition, but we need to handle them in order to read files. So I simply added Some(0) in addition to None for primitive fields and reset of the repetition type on the root node.

Relates to #178.


sunchao:
Merged. Thanks @sadikovi !

sadikovi:
Thanks!

PR #185, opened 2018-11-05
Arrow schema converter.

liurenjie1024:
This is the first step of adding an arrow reader and writer for parquet-rs. 
This commit contains a converter which converts parquet schema to arrow schema.


sunchao:
Thanks @liurenjie1024 . Will take a look soon. 

Meanwhile, I created #186 to track the overall progress of adding Arrow support. Please feel free to create more tasks in it. I'm also planning to take up some tasks. :)

sunchao (commented):
Thanks @liurenjie1024 . Left some comments.

src/arrow_format/schema.rs
```diff
@@ -0,0 +1,965 @@
...
+              Some(DataType::Struct(fields))
+            }
+          })
+      }
+    }
+  }
+  
+  fn to_list(&self) -> Result<Option<DataType>> {
+    match &*self.schema {
+      Type::PrimitiveType {..} => panic!("This should not happen."),
```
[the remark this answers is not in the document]
liurenjie1024: - I think panic when it's primitive type is same.
- I don't think we should panic when the `fields.len() != 1` because this is an error format and should not cause our function to panic.

src/arrow_format/schema.rs
```diff
@@ -0,0 +1,965 @@
...
+            basic_info: _,
+            fields
+          } if fields.len()>1 => {
+            item_converter.to_struct()
+          },
+          Type::GroupType {
+            basic_info: _,
+            fields
+          } if fields.len()==1 && list_item.name()!="array" &&
+            list_item.name()!=format!("{}_tuple", self.schema.name()) => {
```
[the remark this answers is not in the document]
liurenjie1024: They are not the same. The list item name must be equal to `{schema.name()}_tuple`

src/arrow_format/schema.rs
```diff
@@ -0,0 +1,965 @@
...
+
+/// Convert parquet schema to arrow schema.
+pub fn parquet_to_arrow_schema(parquet_schema: SchemaDescPtr) -> Result<Schema> {
+  parquet_to_arrow_schema_by_columns(parquet_schema.clone(), 0..parquet_schema.columns().len())
+}
+
+/// Convert parquet schema to arrow schema, only preserving some leaf columns.
+pub fn parquet_to_arrow_schema_by_columns<T>(
+  parquet_schema: SchemaDescPtr, column_indices: T) -> Result<Schema>
+  where T: IntoIterator<Item=usize> {
```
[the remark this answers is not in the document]
liurenjie1024: So that the user can just pass a vec![1,2,3] without calling .iter(). BTW, if T implements Iterator, it also implements IntoIterator.

liurenjie1024:
@sunchao Thanks for the review and I've fixed the comments. As for the format part, how do you think about a rustfmt file so that other contributors can follow?

src/arrow_format/schema.rs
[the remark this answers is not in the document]
sunchao: You're right. For the latter maybe we can throw "not yet supported" error.

src/arrow_format/schema.rs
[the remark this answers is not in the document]
sunchao: I see. Looks good.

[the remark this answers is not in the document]
liurenjie1024: Yes, I didn't notice the check you mentioned before. I think we need another refactor so that we can share the code.

src/arrow_format/schema.rs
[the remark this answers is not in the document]
sadikovi: Why can't we return Err here?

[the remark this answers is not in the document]
liurenjie1024: Because if this happens, it means our code has bug.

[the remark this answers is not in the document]
sadikovi: To start with, returning an Err does not necessarily mean the code has bug. Err is just a way of expressing an invalid state or a bug and returning it without panicking. With Err it is easier to add a unit test to show that we return a proper error message.

By the way, there is no test for this, either with panic or Err. Also, what does it mean "It should not happen". What should not happen? Can you update the error message?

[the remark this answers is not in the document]
sadikovi: If you want to go with panic, I would suggest including something like the actual schema that was compared, and some other metadata as part of the error message, so it is easier to debug later on when it happens.

src/errors.rs
```diff
@@ -52,6 +52,12 @@ quick_error! {
       display("EOF: {}", message)
       description(message)
     }
+    /// Arrow error.
```
sadikovi: Can you add new line here?

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+    }
+  }
+  
+  fn is_repeated(&self) -> bool {
+    let basic_info = self.schema.get_basic_info();
+    
+    basic_info.has_repetition() && basic_info.repetition()==Repetition::REPEATED
+  }
+  
+  fn is_self_included(&self) -> bool {
```
sadikovi: Why do you need this function? How could self be part of self.leaves?

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+
+/// Convert parquet schema to arrow schema.
+pub fn parquet_to_arrow_schema(parquet_schema: SchemaDescPtr) -> Result<Schema> {
+  parquet_to_arrow_schema_by_columns(parquet_schema.clone(),
+                                     0..parquet_schema.columns().len())
+}
+
+/// Convert parquet schema to arrow schema, only preserving some leaf columns.
+pub fn parquet_to_arrow_schema_by_columns<T>(
+  parquet_schema: SchemaDescPtr, column_indices: T) -> Result<Schema>
```
sadikovi: `) -> Result<Schema>` normally goes onto new line.

src/arrow/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+//! data.
+//!
+//! This mod provides API for converting between arrow and parquet.
+
+pub mod schema;
+
+pub use self::schema::parquet_to_arrow_schema;
+pub use self::schema::parquet_to_arrow_schema_by_columns;
+
+
```
sadikovi: Empty lines.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ lz4 = "1.23"
 zstd = "0.4"
 chrono = "0.4"
 num-bigint = "0.2"
+arrow="0.11.0"
```
sadikovi: Spaces around `=` - you have an example just above this line!

src/arrow/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! [Apache Arrow](http://arrow.apache.org/) is a cross-language development platform for in-memory
+//! data.
+//!
+//! This mod provides API for converting between arrow and parquet.
```
sadikovi: I would probably add a doc test here, showing how to use the function. Optional, but would be nice to have.

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+      .filter_map(|f| f)
+      .collect::<Vec<Field>>())
+    .map(|fields| Schema::new(fields))
+}
+
+/// This struct is used to group methods and data structures used to convert parquet
+/// schema together.
+struct ParquetTypeConverter {
+  schema: TypePtr,
+  leaves: Rc<HashSet<*const Type>>
```
sadikovi: Why is `leaves` an `Rc<HashSet>`?

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+  }
+  
+  // Utility functions.
+  
+  fn is_nullable(&self) -> bool {
+    let basic_info = self.schema.get_basic_info();
+    if basic_info.has_repetition() {
+      match basic_info.repetition() {
+        Repetition::OPTIONAL => true,
+        Repetition::REPEATED => true,
```
sadikovi: Is repeated field nullable? Normally it is complemented by nested optional or required fields. I am not very familiar with arrow, maybe it treats them as nullable by default.

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+      }
+    } else {
+      false
+    }
+  }
+  
+  fn is_repeated(&self) -> bool {
+    let basic_info = self.schema.get_basic_info();
+    
+    basic_info.has_repetition() && basic_info.repetition()==Repetition::REPEATED
```
sadikovi: `basic_info.repetition() == Repetition::REPEATED`.

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+      other => Err(ArrowError(
+        format!("Unable to convert parquet logical type {}", other)))
+    }
+  }
+  
+  // Functions for group types.
+  
+  fn to_struct(&self) -> Result<Option<DataType>> {
+    match self.schema.as_ref() {
+      Type::PrimitiveType {..} => panic!("This should not happen."),
```
sadikovi: Please, extend these error messages, they should indicate the error and potentially include metadata that could aid debugging.

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+  use basic::Repetition;
+  use schema::types::PrimitiveTypeBuilder;
+  
+  use arrow_lib::datatypes::Field;
+  use arrow_lib::datatypes::DataType;
+
+  use super::parquet_to_arrow_schema;
+  use super::parquet_to_arrow_schema_by_columns;
+  
+  macro_rules! make_parquet_type {
```
sadikovi: Could you also check test utils? We might already have a function there to generate schemas.

src/arrow/schema.rs
```diff
@@ -0,0 +1,992 @@
...
+use arrow_lib::datatypes::Schema;
+use arrow_lib::datatypes::Field;
+use arrow_lib::datatypes::DataType;
+
+use errors::ParquetError::ArrowError;
+use errors::Result;
+
+/// Convert parquet schema to arrow schema.
+pub fn parquet_to_arrow_schema(parquet_schema: SchemaDescPtr) -> Result<Schema> {
+  parquet_to_arrow_schema_by_columns(parquet_schema.clone(),
```
sadikovi: We normally format it like this:
```rust
func(
  arg1,
  arg2,
  ...
)
```

src/schema/types.rs
```diff
@@ -718,15 +718,26 @@ impl SchemaDescriptor {
 
   /// Returns column root [`Type`](`::schema::types::Type`) for a field position.
   pub fn get_column_root(&self, i: usize) -> &Type {
+    let result = self.column_root_of(i);
+    assert!(result.is_some(), "Expected a value for index {} but found None", i);
+    result.unwrap().as_ref()
+  }
+
+  /// Returns column root [`Type`](`::schema::types::Type`) pointer for a field position.
+  pub fn get_column_root_ptr(&self, i: usize) -> TypePtr {
```
sadikovi: I think it would be good to move the assertion into `column_root_of` method and return either Rc directly or ref extracted from Rc. In this case we do not need to duplicate the error message.

sadikovi (commented):
Thanks for the PR. I left a few comments. Could you also add doc for the functions even if they are trivial, it is unclear what some of them do and why they do it, and will be difficult to maintain the code later.

src/arrow/schema.rs
liurenjie1024: This function is used to test if the schema included in the leaves. We need it when the converter is converting a primitive schema. The leaves are not leaves of the schema, but columns that need to convert.

src/arrow/schema.rs
liurenjie1024: I think the name `leaves` is quite misleading. It's not the leaves of schema, but the leave columns that needs to convert. I'll change that.

src/arrow/schema.rs
liurenjie1024: Yes, you're right. In fact there's no clear definition about the mapping to arrow and I just think this is default case.

src/arrow/schema.rs
liurenjie1024: If the repetition is not defined, this may cause panic.

sadikovi: I meant white spaces around it. I know that repetition unwraps the option.

src/arrow/schema.rs
liurenjie1024: Thanks. I found a parse which can convert texts to type. I'll change tests to use that.

liurenjie1024:
@sadikovi Sorry for late reply. I've fixed some comments.

liurenjie1024:
@sunchao @sadikovi Could you help to review this?

sunchao:
@liurenjie1024 sure, will take a look soon. Sorry for the delay.

src/arrow/mod.rs
sadikovi: Okay, but let's not forget to do that.

sadikovi (commented):
Looks good.

sunchao (approved):
Sorry for the late review (was on vacation). LGTM.

liurenjie1024:
@sunchao @sadikovi We have two LGTM, help to merge this?

liurenjie1024:
Yes, I think it would be better to merge it after merging with arrow.

sunchao:
Great. We can create an umbrella JIRA after the merge, to track the parquet-arrow integration. I can help to get this PR in very soon after that.

liurenjie1024:
Cool.

sunchao:
@liurenjie1024 I created a JIRA here: https://issues.apache.org/jira/browse/ARROW-4060. Could you file a PR in the arrow repo for this? Thanks.

sunchao:
Closing this as ARROW-4060 is already merged into Arrow.

PR #189, opened 2018-11-10
Apply rustfmt to the code base

### LINKED ISSUE #187: Apply rustfmt?
This has been discussed long time back (see #44). As we have more people contributing to the project now, I wonder if we should revisit this, provide a rustfmt template and check it in CI. Personally, I had some good experiences with rustfmt in other projects, as well as similar stuff in Go. @sadikovi : want to hear you opinion on this.

sunchao:
This applies rustfmt on the codebase with the rules defined in `rustfmt.toml`. 

Fixes #187.


.travis.yml
```diff
@@ -35,10 +35,12 @@ rust:
 
 before_script:
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
+  - rustup component add rustfmt-preview
   - cargo install cargo-travis || echo "Skipping cargo-travis installation as it is already cached"
   - export PATH=$HOME/.cargo/bin:$PATH
 
 script:
+  - cargo fmt --all -- --check
```
sadikovi: Would it better to move format check after running all of the tests?

sunchao: Yes, we could - though it won't make much difference since if test fails the format check won't run.

sadikovi: Fair enough. One could always run tests or format locally. I was thinking if it would be a good idea to add a paragraph with a command to run format. Either way - LGTM.

PR #196, opened 2018-11-28
Add EncodingWriteSupport for column writer

### LINKED ISSUE #195: ParquetDecodingException When Writing Boolean
I have a schema with a field of type boolean defined

```
let message = "message schma {
REQUIRED BOOLEAN a;
}";
```

I am using the BoolColumnWriter to write the boolean data:

```
match col_writer {
     ColumnWriter::BoolColumnWriter(ref mut typed_writer) => {
      typed_writer.write_batch(
       &[true],
        None,
        None
        ).unwrap();
         },
         _ => { }
 }
```

When I try to run the generated parquet file, I get the following exception: 
**org.apache.parquet.io.ParquetDecodingException: Can not read value at 0 in block -1 in file**

I cant seem to get passed this issue. Any idea why this is happening?

sadikovi:
This PR addresses the issue when we would enable dictionary encoding for boolean values regardless of the parquet format v1 or v2. This prevents file to be read by other systems that use parquet-mr. 

In parquet-mr column writer use either PLAIN (v1) or RLE (v2) when writing for boolean values. In our library we would just use an alternative encoding that was set for a column, which defaults to PLAIN.

This PR changes the behaviour. Now if no encoding explicitly set for a column, we will use EncodingWriteSupport trait to check if column type supports dictionary encoding and what alternative fallback encoding we can use. The values match parquet-mr implementation for both V1 and V2. It also changes API for WriterProperties - now `encoding(column)` returns None, if no encoding is set (used to return default PLAIN encoding).

Closes #195


sunchao:
@sadikovi can you update the PR with rustfmt?

sadikovi:
I'd like to discuss the approach:
- I added trait to handle dictionary enabled or disabled. It could be an overkill since we can just use a simple method for it, though trait does make it easy to check the value type.
- I decided to silently fall back to an alternative encoding when the type is boolean. My other approaches included returning Result<ColumnWriter>, which would become Err if boolean has dictionary encoding, or panicking. The first one requires API changes, and the second is just not nice.

Let me know if you have any other suggestions or objections to this approach. Thanks!

sunchao:
@sadikovi 
1. the added trait looks fine to me. I think it clearly expresses the idea that we want to handle boolean type specially. Replacing it with a function sounds good to me as well but I don't have strong preference on either.
2. I'm fine with the silent fallback. We should definitely not panic. Returning `Result` seems not nice too and it's also not quite compatible with how `properties` are implemented at the moment.

I noticed another issue though - it seems we use `PLAIN_ENCODING` as the default fallback encoding for all encoding types, which may not be the best choice. We may want to use different encodings based on types and parquet version (i.e., v1 or v2). This is how parquet-mr is doing.

sadikovi:
Yes, I was thinking about it. I can add it as follow up work.

sadikovi:
Actually, I can do all of that here, let me know if I should include default encodings depending on a column type work in this PR.

sunchao:
I think it may be better to open a new PR since these two are not so closely related.

sunchao:
Yes. Sounds good.

sadikovi:
@sunchao I updated the PR with encoding write support. Can you have a look when you have time? Thanks!

sadikovi:
Thanks for merging!

PR #202, opened 2018-12-06
Patch record_reader_10k_collect benchmark that fails with Int96 conversion

### LINKED ISSUE #201: Benchmark record_reader_10k_collect fails with Int96 error display conversion
When running benchmark `record_reader_10k_collect`, it fails with the following error:
```
running 1 test
test record_reader_10k_collect             ... FAILED

failures:

---- record_reader_10k_collect stdout ----
thread 'main' panicked at 'Expected non-negative milliseconds when converting Int96, found -210866803200000', src/record/api.rs:504:7


failures:
    record_reader_10k_collect

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out
```

It relates to the work we did for Int96-timestamp conversion. I think we should solve it somehow or prioritise work on Int96 conversion.

sadikovi:
This PR tries to mitigate the failure in `record_reader_10k_collect` benchmark by passing projection that excludes Int96 column. The error that I got was:
```
running 1 test
test record_reader_10k_collect             ... FAILED

failures:

---- record_reader_10k_collect stdout ----
thread 'main' panicked at 'Expected non-negative milliseconds when converting Int96, found -210866803200000', src/record/api.rs:504:7
```

I simply pass projection on every iteration of the benchmark. **Unfortunately this adds an overhead of parsing the string into schema every time we run benchmark**, but IMHO it is negligible compared to actual reading of records.

I will revert the patch once Int96 conversion works correctly for negative timestamps.

The benchmark runs after the patch.

Closes #201


sadikovi:
How come we did not see the failure earlier? Does CI run benchmarks?

sunchao:
> How come we did not see the failure earlier? Does CI run benchmarks?

CI does run the benchmark. It's at the bottom of the output, and you'll need to expand it to examine. Seems the benchmark started to fail since the int96 code change.

sunchao:
Merged. Thanks @sadikovi !

PR #204, opened 2018-12-08
Update Row API to store TypePtr instead of String

sadikovi:
This PR updates `Row` struct to store `Vec<(TypePtr, Field)>` instead of `Vec<(String, Field)>`. This improves time when reading Parquet files. 

For this I had to change record reader to pass type pointer instead of field name. This unfortunately makes it difficult to test, so I added a macro to generate dummy type with a field name.

Benchmark results:

Before
```
test record_reader_10k_collect             ... bench:  41,689,686 ns/iter (+/- 33,883,503) = 16 MB/s
test record_reader_stock_simulated_collect ... bench: 335,444,189 ns/iter (+/- 36,688,424) = 3 MB/s
test record_reader_stock_simulated_column  ... bench:  14,664,590 ns/iter (+/- 1,444,286) = 87 MB/s
```

After
```
test record_reader_10k_collect             ... bench:  26,658,590 ns/iter (+/- 1,773,580) = 25 MB/s
test record_reader_stock_simulated_collect ... bench: 140,238,171 ns/iter (+/- 6,796,855) = 9 MB/s
test record_reader_stock_simulated_column  ... bench:  15,275,906 ns/iter (+/- 349,117) = 84 MB/s
```


src/record/api.rs
```diff
@@ -139,13 +142,28 @@ impl RowAccessor for Row {
 
 /// Constructs a `Row` from the list of `fields` and returns it.
 #[inline]
-pub fn make_row(fields: Vec<(String, Field)>) -> Row { Row { fields } }
+pub fn make_row(fields: Vec<(TypePtr, Field)>) -> Row { Row { fields } }
+
+impl PartialEq for Row {
+  fn eq(&self, other: &Row) -> bool {
```
sadikovi: Equality is only used in tests, normally user would not run that in the production code.

src/record/api.rs
```diff
@@ -706,6 +724,20 @@ mod tests {
     }};
   }
 
+  // NOTE:
```
sadikovi: I think we should refactor our tests, there is a lot of code duplication here and there.

sadikovi:
Fails to run `$ rustup component add rustfmt-preview`, I tried triggering build over the weekend and today - still the same error.

sunchao:
Yes, there are some issue in the rustup toolchain right now. See the [reddit thread](https://www.reddit.com/r/rust/comments/a4mn9m/rustfmt_unable_to_download_for_nightly_toolchain/). Let me see if we can use a nightly on a particular date to fix this issue.

sunchao:
Sorry @sadikovi I forgot about this PR. Since `parquet-rs` has merged into Apache Arrow, could you file a JIRA and submit a PR for that? Thanks.

sadikovi:
Yes, no problem. I think it would be good to mark this repo as moved or something similar.

sunchao:
Yes. Will update the README for the merge.

